PR #39, opened 2023-11-30
Draft: 26 input checks

loffleraSMH:
Note: This is a draft PR because I’d like to get the `check_input()` function reviewed first before implementing it within other `Rgemini` functions. 

Current version:

- I created a new `check_input()` function in `utils.R` that checks whether function inputs are of correct type. The function can also perform additional (optional) checks based on input type (e.g., for table inputs, can check if required columns exist and whether they are of correct type etc.) – see function documentation and issue description for more details.
- There is an example of how to use this function within the documentation for `check_input` and I also added some unit tests in `tests/testthat/test-check_input.R` that illustrate some scenarios of checks that should/shouldn’t result in error messages.
- I have also applied the function within `readmission` and `daily_census` to illustrate some use cases. Once the function has been reviewed, I’ll add similar checks to other `Rgemini` functions before merging (might need a quick 2nd round of review then).

I focused on checks I've come across in existing `Rgemini` functions/checks I think would be helpful to have. If there are other checks you think would be useful to add, please let me know.


guoyi-smh (commented):
The name of the `options` argument for is a bit confusing as it is used to set global options in baseR. Maybe use `categories` or something else?

R/utils.R
```diff
@@ -202,25 +210,357 @@ find_db_tablename <- function(dbcon, drm_table, verbose = TRUE) {
...
+#'
+#' If an input object can be one of several acceptable types (e.g.,
+#' `data.table` OR `data.frame`), types should be provided as a character vector
+#' (e.g., `argtype = c("data.frame", "data.table")`).
+#'
+#' @param length (`numeric`)\cr
+#' Optional input specifying the expected length of a given input argument
+#' (e.g., use `length = 2` to check if a vector/list contains 2 elements).
+#'
+#' @param options (`character`)\cr
```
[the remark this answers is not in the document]
loffleraSMH: Great point! I just pushed a commit renaming the `options` argument to `categories`.

R/utils.R
```diff
@@ -202,25 +210,357 @@ find_db_tablename <- function(dbcon, drm_table, verbose = TRUE) {
...
+            "user = getPass('Enter user:'), password = getPass('password'))\n",
+            "\nPlease refer to the function documentation for more details."
+          ),
+          call. = FALSE
+        )
+      }
+
+      ## For all other inputs
+    } else if ((any(argtype == "integer") && !all(is_integer(arginput))) ||
+      !any(argtype == "integer") && !any(class(arginput) %in% argtype)) {
```
guoyi-smh, line 434: The check cannot go through in the following cases:
1. input is integer, argtype is "numeric";
2. input is float, argtype is c("numeric", "integer")

code to reproduce the result:
`test <- function(a, ...){
  check_input(a, ...)
}
test(c(1L, 2L), argtype = "numeric")
test(c(1.5, 2.5), argtype = c("numeric", "integer"))
`

guoyi-smh: I think maybe prioritize resolve case 1 as it is more realistic and both `integer` and `double` should be  `numeric`. Case 2 is much less likely to happen.

loffleraSMH: 1. Yes, you are right! Although `class(1L) == "numeric"` returns `FALSE`, for the purpose of this functions, it's probably a good idea to let class integer pass the numeric test. I just made a change adding another condition in the `if` statement to avoid returning an error when the input is an integer and argtype = "numeric" and also added a test scenario for that in the test file. I also added a note about this in the documentation (including a note about `coltypes` where the logic is slightly different). 
2. Right, I think the best way to handle this is to add an explanation in the documentation. Since integer is a special case of numeric, I think checking whether the input is either or is not very meaningful. Let me know if this makes sense.

loffleraSMH:
Thanks @guoyi-smh! I'll create a separate (low-priority) issue to implement any relevant input checks in other functions (I've only applied them to `readmission` and `daily_census` so far).

Issue #9, opened 2023-09-01
Functionalize mapping warning


vaakesan-SMH:
The following warning can be added as a global variable in `utils.R` and referenced in each function it is used (such that it only needs to be updated in once place to be updated everywhere):

```
"\n***Note:***
  The output of this function is based on manual mapping of imaging modalities by a GEMINI Subject Matter Expert.
  Please carefully check mapping coverage for your cohort of interest, or contact the GEMINI team if you require additional support.\n"
```

vaakesan-SMH:
This can also be added to the `mlaps` function.

PR #40, opened 2023-12-01
Migrate mlaps

vaakesan-SMH:
Needs to be checked against the code review checklist.

Handling of `suppressWarnings` for `min/max(c())` `Inf/-Inf` should ideally also be improved.


vaakesan-SMH:
Note that although I've added a utility function (`return_hospital_field()`) to find the hospital field name, this is a hard check on the `admdad` table.

Ideally we'd use `find_db_tablename()` to find the name of the `admdad` table (some HPC datacuts use a different name for this table). However, I tried this implementation and tested it in `report_db_v3`, but there were multiple matches (`admdad` and `admdad_temp3333`), which caused an error.

Since this function will be primarily used with `report_db`, I've fixed the table to `admdad`. However, I wonder if we could make a change to `find_db_tablename()` so we can take a particular table if multiple are returned? For now this solution will be fine because HPC users will not necessarily need `loop_mlaps()`. Let me know your thoughts @gemini-wenb @loffleraSMH

gemini-wenb:
Good point. The current `find_db_tablename` doesn't handle multiple matches when the matches are unique from each other. 

@loffleraSMH do you know whether /how often that HPC users would have multiple datasets with names starting on the same string (i.e. `admdad` and `admdad_temp3333`)?

I wonder if multiple matches only happens when we search on the entire non-HPC dbs (e.g. `report_db_v3`). In this case, we can add a flag to `find_db_tablename()` for internal use to always return `admdad`. 

This wouldn't work if HPC users can have multiple admdads because we won't be able to hard code which one should be used without knowing the specifics diff in their naming variations.

R/mlaps.R
```diff
@@ -0,0 +1,394 @@
...
+      }
+    )
+  ) %>%
+    as.data.table()
+
+  hospital_years <- admdad %>%
+    select(hospital_id, year) %>%
+    unique()
+
+  res <- purrr::map2_df(
```
vaakesan-SMH, line 136: The `purrr` import is missing at the package level. Needs to be added to the `Imports` in the `DESCRIPTION` file.

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+#' Consider lab tests collected **up to** `hours_after_admission` hours after inpatient admission in the calculation.
+#' Default `hours_after_admission` is set to 0, where only lab tests collected at Emergency Department (before inpatient admission) is considered in mLAPS calculation.
+#' Since not all encounters are admitted through Emergency Department, depending on research question, it can be relevant to consider lab tests collected in early inpatient admission.
+#' Typically, `hours_after_admission` can be set to 24 to consider any lab tests collected at Emergency Department and 24 hours after inpatient admission.
+#'
+#' @param output_laps_components (`logical`)\cr
+#' Does not aggregate the score and instead outputs for each LAPS component (test), its contribution
+#' to the overall score.
+#'
+#' @return (`data.frame`)\cr
```
loffleraSMH, line 93: Maybe add here that patients without entry in the lab table are not returned in the output, so need to be merged with `cohort` again and their `mLAPS` value will be `NA`. Also worth mentioning that lab availability per `genc_id` partially depends on `hours_after_admission` (e.g., `genc_id` may have an entry in the lab table, but not within specified time window`.
Maybe this is too much here, but I'm wondering if we could also add something like this: "Note that in certain scenarios, it might be appropriate to impute missing LAPS with 0. For example, when `hours_after_admission` is set to 24, but some `genc_ids` did not have any lab tests performed within 24h of admission, it might be reasonable to assume that no lab tests were indicated (i.e., LAPS value should be close to 0). However, this assumption can only be made for patients who were hospitalized during time periods and hospitals where lab data were in principle available. Additionally, this assumption may not hold true for all patient cohorts, so researchers should use caution when imputing missing LAPS.

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+  )
+
+  return(res)
+}
+
+
+#' @title
+#' mLAPS
+#'
+#' @details
```
loffleraSMH, line 179: Not sure what the best spot for this is, but can you add somewhere in the documentation how the function handles scenarios where multiple results are found for the same lab tests (looks like the max result is used except for creatinine?)?

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+  return(res)
+}
+
+
+#' @title
+#' Calculates minima when input vector is not empty, else returns NA.
+#'
+#' @details
+#' This is a helper function to suppress default warning message from `base::min()`` function
+#' when all elements in the input vector is NA, which can be problemetic for unit testing.
```
loffleraSMH, line 366: Typo in "problematic".

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+min_result_value <- function(x) {
+  ifelse(all(is.na(x)), NA, min(x, na.rm = TRUE))
+}
+
+#' @title
+#' Calculates maxima when input vector is not empty, else returns NA.
+#'
+#' @details
+#' This is a helper function to suppress default warning message from `base::max()`` function
+#' when all elements in the input vector is NA, which can be problemetic for unit testing.
```
loffleraSMH, line 384: Typo in "problematic".

tests/testthat/test-mlaps.R
```diff
@@ -0,0 +1,171 @@
+# Helper function to create dummy lab data needed for mLAPS unit tests
+dummy_lab <- function(id, omop, value, unit, mintime){
```
loffleraSMH, line 2: Nice! Should we actually add this to `dummy_data.R`? I know this may currently not be useful for other functions, but it might be in the future and is probably easier to find in the same spot as the other `dummy_data` functions.

tests/testthat/test-mlaps.R
```diff
@@ -0,0 +1,171 @@
...
+    test_type_mapped_omop = omop,
+    result_value = value,
+    result_unit = rep(unit, length(value)),
+    collection_date_time =  format(as.POSIXct(mintime, tz = "UTC") + sample(0:(24*60*60 - 1), size=length(value), replace = TRUE),"%Y-%m-%d %H:%M")
+  )
+  return(res)
+}
+
+# Helper function to create dummy admdad data needed for mLAPS unit tests
+dummy_admdad <- function(id, admtime){
```
loffleraSMH, line 14: Note that branch `14_plotting_funcs` has a `dummy_ipadmdad` function, which is a bit more general (see here: https://github.com/GEMINI-Medicine/Rgemini/blob/14_plotting_funcs/R/dummy_data.R). You could copy and paste the code from branch 14 into `dummy_data.R` and then use that - it might be useful to have that function on develop anyway (although it hasn't been reviewed...). Anyway, up to you - it it's too much trouble to switch it over at this point, feel free to leave it as is.

tests/testthat/test-mlaps.R
```diff
@@ -0,0 +1,171 @@
...
+  res <- data.table(
+    genc_id = id,
+    admission_date_time = format(as.POSIXct(admtime, tz = "UTC"),"%Y-%m-%d %H:%M")
+  )
+  return(res)
+}
+
+####### test 1
```
loffleraSMH, line 22: Note that all tests currently fail because there is no input argument called `hours_offset` (should be called `hours_after_admission`).

loffleraSMH (commented):
This looks great, thanks @gemini-wenb and @vaakesan-SMH for cleaning up this function! I've added some comments - mostly minor things that should hopefully be quick to resolve. Let me know if you want to discuss any of my comments. I mostly focused on reviewing documentation and testing usability of the function (i.e., I didn't review the actual calculations). I might do a quick 2nd round of review while you are implementing the current suggestions so I haven't approved the pull request quite yet.

I saw your note about "Handling of suppressWarnings for min/max(c()) Inf/-Inf should ideally also be improved" in the pull request. Let me know if you want me to help out with/review any additional changes you might make to this. More generally, the function currently returns a lot of warnings. Could it be useful to suppress all the `NAs introduced by coercion` warnings to avoid too much clutter in the output? 

Speaking of warnings: Maybe we could also add a warning message in the output mentioning that mLAPS calculations depend on mapping by a GEMINI SME. No need to change it on this branch, but maybe just add a note in issue #9 mentioning that we should show this warning in the `mlaps` function. 

Also, don't forget to deprecate the `mlaps` function from `GEMINIpkg` and move the code into the `deprecate_Rgemini.R` file (sorry if you already did it and I missed it!).

tests/testthat/test-mlaps.R
vaakesan-SMH: I'd like to wait until that branch is merged and then make the change (perhaps we can create an issue for this) to avoid potential merge conflicts?

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+#'
+#' @description
+#' A wrapper around the `mlaps()` function which breaks down the calculation of mLAPS on a
+#' large number of encounters by hospital-year. This avoids memory issues that can be caused
+#' by loading large chunks of the lab table.
+#'
+#' @param db (`DBIConnection`)\cr
+#' RPostgres DB connection.
+#'
+#' @param cohort (`numeric`)\cr
```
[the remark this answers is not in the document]
vaakesan-SMH: Understood. Changing this for consistency, although I believe the preference is generally to use the lowest level data structure possible for function arguments (do we support tibbles? matrices?).

However it's low priority and better to be consistent for now so made the change. Might revisit later.

tests/testthat/test-mlaps.R
loffleraSMH: Sounds good! I'll create a new issue and add the `dummy_ipadmdad()` function to a new branch (don't want to merge the rest of the code on branch 14 yet...). I'll mention that the mlaps test code should be updated accordingly once the `dummy_ipadmdad()` function has been reviewed.

tests/testthat/test-mlaps.R
vaakesan-SMH: Thanks for catching this!

R/utils.R
```diff
@@ -224,3 +226,32 @@ find_db_tablename <- function(dbcon, drm_table, verbose = TRUE) {
...
+#' @param db (`DBIConnection`)\cr
+#' RPostgres DB connection.
+#'
+#' @return (`character`)\cr
+#' `hospital_id` or `hospital_num`, with preference given to `hospital_id` if it exists.
+#'
+return_hospital_field <- function(db) {
+
+  admdad <- find_db_tablename(db, "admdad", verbose = FALSE)
+  fields <- dbListFields(db, admdad)
```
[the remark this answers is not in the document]
vaakesan-SMH: Will opt for the first solution of querying the information schema directly

R/mlaps.R
```diff
@@ -0,0 +1,395 @@
...
+#' A numeric vector of `genc_ids` to restrict the calculation to.
+#'
+#' @param hours_after_admission (`numeric`)\cr
+#' Consider lab tests collected **up to** `hours_after_admission` hours after inpatient admission in the calculation.
+#' Default `hours_after_admission` is set to 0, where only lab tests collected at Emergency Department (before inpatient admission) is considered in mLAPS calculation.
+#' Since not all encounters are admitted through Emergency Department, depending on research question, it can be relevant to consider lab tests collected in early inpatient admission.
+#' Typically, `hours_after_admission` can be set to 24 to consider any lab tests collected at Emergency Department and 24 hours after inpatient admission.
+#'
+#' @param output_laps_components (`logical`)\cr
+#' Does not aggregate the score and instead outputs for each LAPS component (test), its contribution
```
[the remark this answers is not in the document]
vaakesan-SMH: Changing to `component_wise` for consistency

vaakesan-SMH:
Thanks @loffleraSMH for the review. Pushed a commit with all the suggestions implemented. Please have a look and let me know if anything was missed (and re-test the function if possible).

vaakesan-SMH:
> I saw your note about "Handling of suppressWarnings for min/max(c()) Inf/-Inf should ideally also be improved" in the pull request. Let me know if you want me to help out with/review any additional changes you might make to this. More generally, the function currently returns a lot of warnings. Could it be useful to suppress all the `NAs introduced by coercion` warnings to avoid too much clutter in the output?

Missed this comment in my commit. I think that all `coercion to NA` warnings remaining should be coming from `laps_assign_test()`, which more generally needs a bit of a cleanup/refactor - but I am leaning towards moving that piece to a separate issue. @gemini-wenb thoughts?

gemini-wenb:
> > I saw your note about "Handling of suppressWarnings for min/max(c()) Inf/-Inf should ideally also be improved" in the pull request. Let me know if you want me to help out with/review any additional changes you might make to this. More generally, the function currently returns a lot of warnings. Could it be useful to suppress all the `NAs introduced by coercion` warnings to avoid too much clutter in the output?
> 
> Missed this comment in my commit. I think that all `coercion to NA` warnings remaining should be coming from `laps_assign_test()`, which more generally needs a bit of a cleanup/refactor - but I am leaning towards moving that piece to a separate issue. @gemini-wenb thoughts?

Yes, the handling of supresswarning for min/max in the original comment has been handled. Remaining warnings is not due to min/max. I also think we should move `laps_assign_test()` cleanup as a separate issue.

tests/testthat/test-mlaps.R
loffleraSMH: Great, thanks - test are all running fine now.

vaakesan-SMH:
> Thanks @vaakesan-SMH for making all those changes. Looks great! I just made a small fix to `return_hospital_field()` (see above). The only other thing I noticed is that the function takes forever to run on `drm_cleandb_v2_1_1` (likely due to FDW, it's fine on `drm_cleandb_v1_2_2`), but we can leave it as is for now as that shouldn't be an issue in the future.
> 
> I approved the pull request so feel free to go ahead and merge now.

Right that's a good point. We actually had this discussion with Trong. `mlaps` will still be calculated by the DB team and available in `derived_variables`, and will generally be run on `report_db` before creating a `drm_cleandb`. For `HPC` datacuts I believe the FDW isn't an issue, is that correct?

The only issue would be if someone wants to run a custom configuration of `mlaps` (`component_wise` or `hours_after_admission`) on their cohort, but really in that case they could just use the `mlaps` function directly instead of the `loop_mlaps` wrapper. The wrapper was basically designed for the DB team to run the function on the entirety of `report_db`.

loffleraSMH:
> > Thanks @vaakesan-SMH for making all those changes. Looks great! I just made a small fix to `return_hospital_field()` (see above). The only other thing I noticed is that the function takes forever to run on `drm_cleandb_v2_1_1` (likely due to FDW, it's fine on `drm_cleandb_v1_2_2`), but we can leave it as is for now as that shouldn't be an issue in the future.
> > I approved the pull request so feel free to go ahead and merge now.
> 
> Right that's a good point. We actually had this discussion with Trong. `mlaps` will still be calculated by the DB team and available in `derived_variables`, and will generally be run on `report_db` before creating a `drm_cleandb`. For `HPC` datacuts I believe the FDW isn't an issue, is that correct?
> 
> The only issue would be if someone wants to run a custom configuration of `mlaps` (`component_wise` or `hours_after_admission`) on their cohort, but really in that case they could just use the `mlaps` function directly instead of the `loop_mlaps` wrapper. The wrapper was basically designed for the DB team to run the function on the entirety of `report_db`.

I'm honestly not sure about HPC (I think the datacuts use FDWs too?), but yes, great point about running `mlaps` function directly without wrapper when re-deriving mLAPS for smaller cohorts.  So yes, I think that's good for now!

Issue #35, opened 2023-11-24
Table1: Different handling of digits for continuous vs. categorical variables


loffleraSMH:
Please identify the affected function: 

`cell_suppression.R` for `table1`

The renders for categorical vs. continuous variables currently handle the `digits` argument differently. For continuous variables, there seems to be a bug where the digits before the decimal point are included in the digit count. For categorical variables, the behaviour is as expected (digits after decimal point). Here is an example:

```
set.seed(999)
gender <- c(sample(c("M", "F"), size = 95, replace = TRUE), rep(NA, times = 5))
age <- rnorm(100, mean = 70, sd = 5)

data <- data.frame(gender = gender, age = age)

tab1 <- table1(~gender + age,
               data = data,
               render.categorical = render_strict_cell_suppression.categorical,
               digits = 5)

tab1
```
![Screenshot 2023-11-24 153935](https://github.com/GEMINI-Medicine/Rgemini/assets/104941195/a112fe9c-f2ac-4b32-bb15-d27e33778ade)

*For the developer:* Once the bug is resolved, please add the reproducible example above as a test in the `tests/testthat` directory.

Any suggested developers?

* 

Any suggested reviewers?

*

vaakesan-SMH:
Needs to change `sprintf` format in `render_cell_suppression.categorical()` and `render_strict_cell_suppression.categorical()` from:

```
paste0("%d (%0.", args$digits, "f%%)")
```

to

```
sprintf(paste0("%d (%.", args$digits, "g%%)"), frequency, pct)
```

To use "significant digits" rather than "digits after the decimal".

loffleraSMH:
@vaakesan-SMH Oh I see, so you would change how categorical variables are handled? I thought it's more intuitive to count digits after the decimal point (i.e., leave categorical as is, but change the behaviour for continuous variables).

vaakesan-SMH:
Agree it's more intuitive to count digits after the decimal point. This is only an issue if the user is "mixing and matching" between `table1` native render functions and `Rgemini` render functions:

```
set.seed(3)

exposure <- sample(
  c("pre-pandemic", "pandemic", "post-pandemic"), 
  size = 100, replace = TRUE
)

age <- rnorm(100, mean = 70, sd = 5)

condition <- sample(
  c("DVT", "CVD", "DM", "Pneumonia", "Dementia"), 
[...]

PR #46, opened 2023-12-06
Add nuances to rounding in cell suppression vignette

vaakesan-SMH:
See discussion in #35.

The chosen solution is to more thoroughly document the edge cases of rounding behaviour when mixing and matching `table1` render functions with `Rgemini` render functions.

Vignette can be knit and reviewed. The knitted markdown is already added to the `.gitignore`.


loffleraSMH (approved):
This is great, thanks @vaakesan-SMH for updating the vignette. The additional explanation is super helpful and very clear. Happy for this to be merged.

PR #48, opened 2023-12-07
Adds function to calculate Elixhauser comorbidity index

### LINKED ISSUE #4: Elixhauser Comorbidity Score
Enable the optional calculation of elixhauser comorbidities scores as per https://pubmed.ncbi.nlm.nih.gov/19433995/

Sample code:

```
elixhauser_comorbidities <- comorbidity(
  x = diagnoses_at_admission,
  id = "genc_id",
  code = "diagnosis_code",
  map = "elixhauser_icd10_quan",
  assign0 = FALSE
)

elixhauser_score <- score(elixhauser_comorbidities, weights = "vw", assign0 = FALSE)
```

Simply provide a convenient wrapper for this.

vaakesan-SMH:
Closes #4.

This requires some refactoring of the comorbidity calculations so that each of charlson and elixhauser are simple wrappers.  Otherwise, documentation is shared as much as possible.

Small edit to the `.Rproj` file to force redocumentation on package build.


R/charlson_comorbidity_index.R
```diff
@@ -124,67 +123,103 @@ diagnoses_at_admission <- function(ipdiag, erdiag) {
...
 #' @importFrom comorbidity comorbidity score
 #' @export
 #'
-charlson_comorbidity_index <- function(ipdiag, erdiag, at_admission = TRUE, raw_comorbidities = FALSE) {
+comorbidity_index <- function(ipdiag, erdiag, map, weights, at_admission = TRUE, raw_comorbidities = FALSE) {
   diagnoses <- if (at_admission) {
     diagnoses_at_admission(ipdiag, erdiag)
   } else {
     rbind(
       erdiag[, .(genc_id, "diagnosis_code" = er_diagnosis_code, "diagnosis_type" = er_diagnosis_type)],
```
guoyi-smh, line 161: When `at_admission = TRUE`, the `diagnoses_at_admission` coerce the input `ipdiag` and `erdiag` to `data.table`s but when `at_admission = FALSE` this step is missing, and as a result when `at_admission = FALSE` and when either input is `data.frame` the function stops.

R/charlson_comorbidity_index.R
```diff
@@ -194,3 +229,47 @@ charlson_comorbidity_index <- function(ipdiag, erdiag, at_admission = TRUE, raw_
```
guoyi-smh, line 229: When `raw_comorbidities = FALSE` the returned table from `comorbidity_index` does not contain the comorbidity index. This naming part should only run when `raw_comorbidities == FALSE` otherwise it messes the column names in the returned table. Same issue with the `elixhauser_comorbidity_index` function.

code to reproduce the result:
```
cleandb <- dbConnect(odbc::odbc(), 
                     Driver = "PostgreSQL Unicode",
                     Server = "geminiaio.smh.smhroot.net", 
                     Port = "5432",
                     Database = "drm_cleandb_v2_1_1", 
                     UID = "guoyi", 
                     PWD = getPass::getPass("Database password: "))

erdiag <- dbGetQuery(cleandb, "select * from erdiagnosis;")
ipdiag <- dbGetQuery(cleandb, "select * from ipdiagnosis")
test <- charlson_comorbidity_index(ipdiag, erdiag, at_admission = T, raw_comorbidities = T)
```

vaakesan-SMH: Great catch! Should be resolved now. Also resolved an issue where when outputting raw comorbidities, the columns were being dropped with this change.

guoyi-smh (approved):
Tested and looks good! 

One question I am curious: when an encounter is present in the `ipdiag` or `erdiag` table, but by the HMSR definition does not have any `diagnosis_at_admission`, in current version the encounter is excluded from the output when `at_admission = TRUE`, which would make them the same as those without any diagnosis data. Is this the desired results or should they be included and get 0 for their comorbidity index?

R/comorbidity_index.R
```diff
@@ -184,7 +186,14 @@ comorbidity_index <- function(ipdiag, erdiag, map, weights, at_admission = TRUE,
 
   scores <- score(comorbidities, weights = weights, assign0 = TRUE)
 
-  res <- cbind(comorbidities$genc_id, scores) %>% as.data.table()
+  res <- cbind(genc_id = comorbidities$genc_id, scores) %>% as.data.table()
+
+  if (at_admission) {
+    res <- all_diagnoses %>%
+      select(genc_id) %>%
```
guoyi-smh, line 193: Here it should be `select(unique(genc_id))` otherwise there are duplicated rows for each `genc_id` in the output.

guoyi-smh (commented):
If we decide to output `0` on index at admission for cases with diagnosis data but no `diagnosis_at_admission`, to be consistent when `raw_comorbidities = T` the raw comorbidities should include these patients and output `0` for them as well?

vaakesan-SMH:
> If we decide to output `0` on index at admission for cases with diagnosis data but no `diagnosis_at_admission`, to be consistent when `raw_comorbidities = T` the raw comorbidities should include these patients and output `0` for them as well?

That is a good point, I will make this fix.

vaakesan-SMH:
> If we decide to output `0` on index at admission for cases with diagnosis data but no `diagnosis_at_admission`, to be consistent when `raw_comorbidities = T` the raw comorbidities should include these patients and output `0` for them as well?

@guoyi-smh, I've made the change to output 0 for raw comorbidities when no diagnoses at admission are present. Please review and let me know what you think.

guoyi-smh (approved):
Tested.

PR #49, opened 2023-12-11
28 add function to check missingness

guoyi-smh:
Adds `n_missing` function to check missingness. `n_missing` works on both `vector` and `data.frame` or `data.table` and allows user to specify either to return number of missingness or index. `mi2` is kept as an alias of `n_missing`.


R/utils.R
```diff
@@ -34,20 +34,94 @@ NULL
...
+#' be returned. For `data.frame` or `data.table` input, the function returns a
+#' vector with length equal to the number of rows of input table and any
+#' missingness in the row will result in a returned value of `TRUE` for the row.
+#'
+#' @return (`logical`)
+#' For `vector` input, a vector of logical values the same length of `x` where
+#' each value represents whether that particular element is missing.
+#' For `data.frame` or `data.table` input, a vector of logical values where
+#' length equals to number of rows of input table and each value represents
+#' whether any element of the row is missing.
```
vaakesan-SMH, line 63: I'm not sure if this is true? Isn't this outputting a named `character` vector and not a `logical` vector?

vaakesan-SMH (changes requested):
Could you add these two functions to `_pkgdown.yml` so they are added to our `pkgdown` documentation?

Otherwise tested and works well!

R/utils.R
guoyi-smh: It should be `character` or `logical` depending on the input `index` argument. `character` is missed here and I will add it.

vaakesan-SMH:
Just confirming @guoyi-smh that this function now supersedes `GEMINIpkg::count_missing()`? If so we will need to submit a merge request for `GEMINIpkg` that deprecates that function.

guoyi-smh:
> Just confirming @guoyi-smh that this function now supersedes `GEMINIpkg::count_missing()`? If so we will need to submit a merge request for `GEMINIpkg` that deprecates that function.

Yes I think it is safe to deprecate `GEMINIpkg::count_missing()`.

vaakesan-SMH:
> > Just confirming @guoyi-smh that this function now supersedes `GEMINIpkg::count_missing()`? If so we will need to submit a merge request for `GEMINIpkg` that deprecates that function.
> 
> Yes I think it is safe to deprecate `GEMINIpkg::count_missing()`.

I think we will also need to do the same for `mi2()`. Could you create a merge request on `GEMINIpkg` which deprecates these two and assign me as reviewer to that as well?

vaakesan-SMH (approved):
Looks good! Thanks for the implementation @guoyi-smh.

Curious how we handle `NULL`, because it can't be added to the `na_strings` vector. We don't necessarily have to make any changes now, just wondering if we might need to extend the functionality at some point.

guoyi-smh:
> Looks good! Thanks for the implementation @guoyi-smh.
> 
> Curious how we handle `NULL`, because it can't be added to the `na_strings` vector. We don't necessarily have to make any changes now, just wondering if we might need to extend the functionality at some point.

Good point. I did think of this but ended up not implementing it in this version. So far I have not encountered any `NULL` in our database. In `lab` table there are some `"NULL"` as strings so the function works on them if the user specify `na_strings = "NULL"`. And it takes some hack to create a dataframe/data.table with `NULL`. So I think we are good for now. But definitely it will be great to add in the future as an improvement.

vaakesan-SMH:
> Good point. I did think of this but ended up not implementing it in this version. So far I have not encountered any `NULL` in our database. In `lab` table there are some `"NULL"` as strings so the function works on them if the user specify `na_strings = "NULL"`. And it takes some hack to create a dataframe/data.table with `NULL`. So I think we are good for now. But definitely it will be great to add in the future as an improvement.

Makes sense to me. Feel free to go ahead and resolve merge conflicts, merge this branch (squash all commits when merging), and then merge your branch on `GEMINIpkg`.

Thanks @guoyi-smh !

PR #59, opened 2023-12-14
44 covid flag er

### LINKED ISSUE #44: Improve documentation for erdiagnoses in `covid_flag()`
## New Feature Request

Currently, the rationale for when and why to include/exclude ER diagnoses is not very clear in the function documentation. Provide additional explanation and reference for lower validity of ER diagnoses (Surain may know reference). Also see issue #43.

To be discussed: We may want to make a small change to the input arguments for consistency with other functions using diagnosis codes (Charlson Comorbidity, frailty, disability) - which take 2 separate inputs `ipdiag` and `erdiag`. `erdiag` can be set to `NULL` by user if they don't want to include ER diagnoses. Probably want to output a message explaining whether ER diagnoses are included or not, similarly to disability function (see #25).

Note: ED diagnosis documentation also needs to be improved in `charlson_comorbidity` (issue #43) and `disability` functions.

loffleraSMH:
Closes #44 

- Added `erdiag` input, which needs to be explicitly set to `NULL` if users want to exclude ER diagnoses (same as in `disability` and `frailty_score` functions)
- Updated documentation accordingly and clarified scenario when users may want to exclude ER diagnoses (e.g., different rates of ER entry in different cohorts)
- Cleaned up documentation a bit more in general
- Added `check_input` function to check cohort and diagnosis inputs
- Added unit tests: Note that the `dummy_diag` function does not simulate COVID diagnoses (`dummy_diag(nid = 5, nrow = 10, pattern = "^U07")` returns no output), and therefore, the test cases have been hard-coded to include codes "U071" and "U072".


R/covid_flag.R
```diff
@@ -2,34 +2,24 @@
...
-#' Return a dataframe with two separate flags identifying
+#' Returns a table with two separate flags identifying
 #' encounters with confirmed and suspected COVID-19 diagnoses
-#' using ICD-10-CA U071 and U072 respectively
+#' using ICD-10-CA U07.1 and U07.2 respectively.
 #'
-#' A confirmed COVID-19 diagnosis (U071) is coded when
-#' there is a positive COVID-19 test regardless of the the test type.
+#' A confirmed COVID-19 diagnosis (U07.1) is coded when
+#' there is a positive COVID-19 test, regardless of the the test type.
```
gemini-wenb, line 13: What does "regardless of the test type" refer to exactly? 

U071 is initially coded only when it's confirmed by PCR lab test. Only recently that this has been changed to include additional tests (e.g. rapid tests). So this sentence may not reflect the coding practice. I suggest rephrasing it with additional details, or remove this part.

Here is the [CIHI reference](https://www.cihi.ca/sites/default/files/document/mapping-covid-19-codes-icd-10-ca-to-base-icd-10-job-aid-en.pdf ) for change in coding practice for test type: "Change in direction, effective April 1, 2023 — U07.1 COVID-19, virus identified is assigned when COVID-19 is confirmed by a COVID-19 test result. A COVID-19 test may include a polymerase chain reaction (PCR) test, a rapid antigen test (RAT) or a point-of-care molecular test (POCT). Prior to April 1, 2023, U07.1 was assigned only when COVID-19 was confirmed by a COVID-19 lab test."

gemini-wenb (commented):
I reviewed and tested the function. Everything works as expected. I only have one comment on the accuracy of function documentation. Good to merge after that is addressed.

R/covid_flag.R
loffleraSMH: Thanks @gemini-wenb for pointing this out! I've updated the documentation to provide the additional clarification based on the CIHI reference.

PR #64, opened 2024-01-08
functionalized mapping warning #9

vaakesan-SMH:


R/n_routine_bloodwork.r
```diff
@@ -52,15 +52,16 @@
...
 n_routine_bloodwork <- function(dbcon,
                                 cohort) {
   cohort <- coerce_to_datatable(cohort)
 
-  cat("\n***Note:***
-  The output of this function is based on manual mapping of Sodium and Heamoglobin tests by a GEMINI Subject Matter Expert.
-  Sodium and Hemoglobin are prioritized and mapped for all encounters. However, please carefully check mapping coverage for your cohort of interest, or contact the GEMINI team if you require additional support.\n")
+  mapping_message(
+    "Sodium and Hemoglobin tests",
+    "Note however that Sodium and Hemoglobin are prioritized and mapped for all encounters.\n"
```
loffleraSMH, line 63: I think I would just keep the standard message here and remove the additional sentence as it doesn't really add anything.

R/utils.R
```diff
@@ -691,3 +691,30 @@ check_input <- function(arginput, argtype,
...
+#' @param what (`character`)\cr
+#' Which values were mapped.
+#'
+#' @param addtl (`character`)\cr
+#' An additional, specific message to append to the generic message.
+#'
+#' @return
+#' Prints a message to the console.
+#'
+mapping_message <- function(what, addtl = NULL) {
```
loffleraSMH, line 712: Function looks great and works as expected! I like the optional additional input in case we'll need to specify additional details for some mappings.

loffleraSMH (approved):
Thanks @vaakesan-SMH for developing the `mapping_message` function and adding it to the existing functions. Tested it and everything runs as expected. Just one minor comment about the message re bloodwork above. Additionally, I think we can add this to `mlaps` too? `icd_to_ccsr` also has a similar message, although the mapping isn't strictly manual (it's automated, but requires some manual review...). I'll leave it up to you if you want to replace the existing message with the function or leave it as is. Happy for this to be merged as soon as you think it's ready.

R/n_routine_bloodwork.r
vaakesan-SMH: Agree, removed.

vaakesan-SMH:
>I think we can add this to `mlaps` too? `icd_to_ccsr` also has a similar message, although the mapping isn't strictly manual (it's automated, but requires some manual review...).

Decided to add the message to `mlaps` and replace the message in `icd_to_ccsr` (although it is not technically rigorous, it captures the same spirit of why the user needs to be informed).

Issue #74, opened 2024-02-05
Bug in `check_input` when class returns multiple column types


loffleraSMH:
Please identify the affected function:  `check_input()`

The `check_input` function currently fails for checks on data.frames/data.tables where `class(column_name)` returns more than 1 class (e.g., date-times that return class `POSIXct` and `POSIXt`). The problem is in check 6 of `check_input` (line 565) which says `!(coltype, class....)` but should say `!any(coltype, class...)`.

![image](https://github.com/GEMINI-Medicine/Rgemini/assets/104941195/029c9414-f703-4168-81fa-5898952a6588)

PR #75, opened 2024-02-05
74 bug input check

loffleraSMH:
- Closes issue #74 

Fixed the bug described in the issue and added a new unit test to check for the scenario causing the bug in the previous version of the function. 

An easy way to check the old vs. new function is to try running this code - the old version will return an error, whereas the fixed version should run smoothly:
``
check_input(data.table(discharge_date_time = lubridate::ymd_hm('2020-01-01 00:00')),
                c("data.frame", "data.table"),
                colnames = c("discharge_date_time"),
                coltypes = c("POSIXct"))
``


loffleraSMH:
> data <- data.table::data.table(discharge_date_time = lubridate::ymd_hm('2020-01-01 00:00'))
> Rgemini:::check_input(data, c("data.frame", "data.table"), colnames = c("discharge_date_time"), coltypes = c("POSIXt"))

Thanks for reviewing this so quickly @vaakesan-SMH! Oh interesting, both examples fail for me on `master`. But yes, as long as both checks pass on the new version, that's great - so I'll go ahead and merge this.

PR #87, opened 2024-03-07
86 tablename subset fix

### LINKED ISSUE #86: [BUG]: `find_db_tablename()` does not identify `_subset` tables
Please identify the affected function: `find_db_tablename()`

The current version of this function fails to identify tables on HPC4Health due to the fact that datacut tables use the following convention `"tablename_subset"`, however, the following line of code is missing the underscore:
``` 
res <- table_names[table_names %in% c(table, paste0(table, "subset"))]
```

The correct version should be this:
``` 
res <- table_names[table_names %in% c(table, paste0(table, "_subset"))]
```

loffleraSMH:
Closes #86 (see issue description for bug and explanation for solution implemented in this PR)


vaakesan-SMH:
I couldn't actually test this properly because I do not have access to any HPC datacuts that have this "subset" naming convention for some tables. However, from a basic review standpoint this PR is simple enough and looks good, so I'm still comfortable to merge.

Let me know if you'd like me to test more thoroughly.

loffleraSMH:
> I couldn't actually test this properly because I do not have access to any HPC datacuts that have this "subset" naming convention for some tables. However, from a basic review standpoint this PR is simple enough and looks good, so I'm still comfortable to merge.
> 
> Let me know if you'd like me to test more thoroughly.

Thanks @vaakesan-SMH for reviewing! Yes, I know it's a bit tricky to test this, but I think it's fine in this case. I will run a final test once the new version has been moved over to HPC - just to be sure. In the future, maybe we can think of a better way to integrate testing on HPC into our workflow to make sure everything works well before versioning up. For now, I'll go ahead and merge this.

Issue #14, opened 2023-09-07
Add plotting functions for exploratory analyses


loffleraSMH:
## New Feature Request

Adding a first set of functions that facilitate common ways of exploring the data and visualizing some basic stats:

1. Summary plot with basic descriptive histograms/barplots for multiple variables
2. Function plotting a variable over time (and grouped by hospital, or other grouping variables).

Suggested developers:

* @sophialaird 
* @loffleraSMH

shijiaSMH:
@loffleraSMH, adding a couple of ideas for more flexibility here, let's add those if they are quick fixes. If not, let's leave them for next time.

1. Show both scattered dots and line; use simple linear regression to fit a smoother line to see trends better
![image](https://github.com/GEMINI-Medicine/Rgemini/assets/105071952/4b4a935f-d93f-41c3-b391-471ba9190825)

2. Flexibility for line types and dot types
![image](https://github.com/GEMINI-Medicine/Rgemini/assets/105071952/4542c4bc-6759-40bf-bf14-cae66d92d5c5)
![image](https://github.com/GEMINI-Medicine/Rgemini/assets/105071952/99132594-84d6-435e-9e42-7eea483b1892)

3. Flexibility to add text/label
![image](https://github.com/GEMINI-Medicine/Rgemini/assets/105071952/fbeeb497-21b8-476f-a4cf-c4d81a729d1b)

loffleraSMH:
@shijiaSMH: Hm, let me think about whether we can add 1 (but I might open a new issue for this). Not sure about 2 (we could add a simplified version of this but this is getting into very user-specific needs that might make the function unnecessarily complex for a basic exploratory/descriptive plot). 3. can simply be achieved by users adding any additional layers to the plot with `+` so no need to change the function for this.

shijiaSMH:
sorry just added 4. time cut, can you see?

Cool just don't want any inflexibility for users to not use those functions!

loffleraSMH:
4. can also be added by users with a single line of code, once the basic plot was returned by the function, correct?
Yep, let's make sure the basic versions of these functions work properly first (there is already lots of potential for bugs) and then we can add other enhancements (based on demand) later on?

loffleraSMH:
[...]

PR #113, opened 2024-05-08
14 plotting funcs

### LINKED ISSUE #112: Add plotting theme & colors
## New Feature Request

Create simple, clean theme that can be applied to any ggplot figures, and will be used as default theme for all Rgemini plotting functions.

Let's also add some basic color palettes to have a convenient way to pick some basic color options.

Note: This issue was previously combined with #14 and all code is implemented on branch 14_plotting_funcs

loffleraSMH:
Closes issues #14 and closes #112

All code is on a single branch (14_plotting_funcs) but I separated the previous "plotting functions" issue into 2 separate issues that can be reviewed independently of each other (because there is a lot of code/documentation to be reviewed). 

- The reviewer for #14 should review the vignette `plotting_data_exploration.Rmd` and the 2 functions covered in that vignette: `plot_summary()` and `plot_over_time`
- The reviewer for #112 should review the vignette `plotting_theme.Rmd` and the 2 functions covered in that vignette: `plot_theme()` and `gemini_colors`

A knitted version of the most up-to-date vignettes is saved here: `R:\GEMINI\Research Projects\Rgemini\`

I'd recommend that reviewers start by going through the vignettes and make sure the rationale & use of the functions is explained clearly. Reviewers should then review the actual function code and test the functions on additional scenarios (e.g., different tables, edge cases like missing data/weird variable types etc.).


R/plot_summary.R
shijiaSMH, line 1: line 75 need to add physician to avoid admitting_physician, discharging_physician? Or I think too much
      !grepl("genc_id|patient_id|cpso|physician|hospital_id|hospital_num|_date|_time"

vignettes/plotting_data_exploration.Rmd
shijiaSMH, line 1: line 85 (ignoring any encounter-/patient-/hospital-ID variables and date-time variables)

you have cpso in your codes, so can add physician ID;
For shortening, can change to: (ignoring any ID-related and date-time variables)

R/plot_summary.R
shijiaSMH, line 1: line 89 & 93: same typo
change to: ## add varlable as list item

loffleraSMH: Good point, just pushed a fix to remove all relevant physician ID variables. 
```
    plot_vars <- colnames(data)[
      !grepl("genc_id|patient_id|hospital_id|hospital_num|_date|_time|
             cpso|physician|adm_code_raw|dis_code_raw|mrp_code_raw",
        colnames(data),
        ignore.case = TRUE
      ) & colnales(data) != "mrp"
    ]
```

vignettes/plotting_data_exploration.Rmd
loffleraSMH: I'll leave it as is to be more specific (users may have other ID variables in their tables that aren't removed here).

R/plot_summary.R
loffleraSMH: This refers to the "varlabel" variable that is being added. Just made a small change so it's less confusing.

shijiaSMH, line 1: Do we care about this warning: `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.?
It appears for: plot_summary(data = ipadmdad)

shijiaSMH, line 1: error: I think this is due to data size too big, do we care about this?

> adm <- dbGetQuery(clean_db, "select * from admdad
                              ") %>% as.data.table
> plot_summary(data = adm)
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
Error in `geom_text()` at Rgemini/R/plot_summary.R:244:8:
! Problem while computing stat.
i Error occurred in the 2nd layer.
Caused by error:
! cannot allocate vector of size 13.6 Mb
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace()
<error/rlang_error>
Error in `geom_text()` at Rgemini/R/plot_summary.R:244:8:
! Problem while computing stat.
i Error occurred in the 2nd layer.
Caused by error:
! cannot allocate vector of size 13.6 Mb
---
Backtrace:
     x
  1. \-global plot_summary(data = adm)
  2.   +-base::suppressWarnings(ggarrange(plotlist = sub_figs, ...)) at Rgemini/R/plot_summary.R:282:2
  3.   | \-base::withCallingHandlers(expr, warning = function(w) invokeRestart("muffleWarning"))
  4.   \-ggpubr::ggarrange(plotlist = sub_figs, ...) at Rgemini/R/plot_summary.R:282:2
  5.     \-purrr::map(...)
  6.       \-ggpubr (local) .f(.x[[i]], ...)
  7.         \-cowplot::plot_grid(plotlist = plotlist, ...)
  8.           \-cowplot::align_plots(...)
  9.             \-base::lapply(...)
 10.               \-cowplot (local) FUN(X[[i]], ...)
 11.                 +-cowplot::as_gtable(x)
 12.                 \-cowplot:::as_gtable.default(x)
 13.                   +-cowplot::as_grob(plot)
 14.                   \-cowplot:::as_grob.ggplot(plot)
 15.                     \-ggplot2::ggplotGrob(plot)
 16.                       +-ggplot2::ggplot_gtable(ggplot_build(x))
 17.                       | \-ggplot2:::attach_plot_env(data$plot$plot_env)
 18.                       |   \-base::options(ggplot2_plot_env = env)
 19.                       +-ggplot2::ggplot_build(x)
 20.                       \-ggplot2:::ggplot_build.ggplot(x)
 21.                         \-ggplot2:::by_layer(...)
 22.                           +-rlang::try_fetch(...)
 23.                           | +-base::tryCatch(...)
 24.                           | | \-base (local) tryCatchList(expr, classes, parentenv, handlers)
 25.                           | |   \-base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
 26.                           | |     \-base (local) doTryCatch(return(expr), name, parentenv, handler)
 27.                           | \-base::withCallingHandlers(...)
 28.                           \-ggplot2 (local) f(l = layers[[i]], d = data[[i]])
 29.                             \-l$compute_statistic(d, layout)
 30.                               \-ggplot2 (local) compute_statistic(..., self = self)
 31.                                 \-self$stat$setup_params(data, self$stat_params)
 32.                                   \-ggplot2 (local) setup_params(..., self = self)
 33.                                     \-ggplot2::resolution(data[[x]], discrete = TRUE)
 34.                                       +-scales::zero_range(range(x, na.rm = TRUE))
 35.                                       \-base::range(x, na.rm = TRUE)
 36.                                         +-base::c(..., recursive = TRUE)
 37.                                         \-ggplot2:::c.mapped_discrete(..., recursive = TRUE)
 38.                                           +-ggplot2:::mapped_discrete(unlist(lapply(list(...), unclass)))
 39.                                           \-base::unlist(lapply(list(...), unclass))
Run rlang::last_trace(drop = FALSE) to see 5 hidden frames.

loffleraSMH: This is generated by `ggplot2::geom_histogram()` and can't be suppressed with `suppressWarnings`. I personally think it's fine to leave it as is.

shijiaSMH: Hi! Thx for adding above, but we still need more tweaks. 

We need to ensure codes don't run for all variables below from physicians table:
> names(phs)
 [1] "genc_id"                   "adm_code_raw"              "dis_code_raw"             
 [4] "mrp_code_raw"              "admitting_physician"       "adm_phy_cpso_mapped"      
 [7] "admitting_physician_gim"   "discharging_physician"     "dis_phy_cpso_mapped"      
[10] "discharging_physician_gim" "mrp"                       "mrp_cpso_mapped"          
[13] "mrp_gim"                   "hospital_id"               "mrp_cpso_hashed"          
[16] "adm_phy_cpso_hashed"       "dis_phy_cpso_hashed"       "hospital_num"   

For now, even if I applied ur updated codes above, I still got plots below (screenshots):
image
image

loffleraSMH: Hm, I'm not getting this error but it raises a good point that I think people may want to avoid running this on large tables without specifying `plot_vars`. I have now added the following warning if users don't provide a `plot_vars` input and the input table contains 10 or more variables to be plotted (ignoring ID/date-time variables): 
`No "plot_vars" input provided for large table. plot_summary() will plot all variables (except ID/date-time variables) in "data" input (= XX variables). This approach is not recommended for large tables with many columns. Instead, we recommend explicitly specifying a subset of variables you want to plot to avoid memory issues/cluttered outputs.`

Do you think that's ok?

loffleraSMH: Can you try pulling again and make sure you source the re-pulled code? This looks like an older version (each plot shouldn't contain more than 9 subplots). I also added another fix just now to abbreviate the x-tick labels to a max of 10 characters. Let me know if it looks better now.

shijiaSMH: yes will pull again and re-test

shijiaSMH: that sounds reasonable, again, will pull & test again!

shijiaSMH: I did repull, still show a coln in physicians, see screenshot

image

also, additional warning: 
Warning message:
```
The `legend.text.align` argument of `theme()` is deprecated as of ggplot2 3.5.0.
i Please use theme(legend.text = element_text(hjust)) instead.
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated. 
```

loffleraSMH: Yes, that's correct - it should show this variable. I'm only removing physician IDs (because there are potentially thousands, so it doesn't make much sense to show barplots).

loffleraSMH: For the warning about `legend.text`: Good catch - I'll push a fix for that shortly.

shijiaSMH, line 1: I think we should do sth to let function stop immediately if user clumsily select an ID/date-time variables. For ex, I ran codes below, was stuck for 10min and function was still running. 

`plot_summary(data = adm, plot_vars = c("discharge_date_time", "number_of_alc_days"))`

shijiaSMH: Just tested, saw those when not selecting variables, closing this thread now

shijiaSMH, line 1: For this ex, this variable has too many values, making values unreadable on x-axis, and some % texts unclear on plots. We can have 2 arguments about this: 1) don't care since we imagine this is a rare use case; 2) add warning to advice ppl not to plot categorical variable exceeding X values OR code change automatically reduce size of text

image

shijiaSMH, line 1: should we add a warning when user try to select both ID/date-time and an eligible variable?
```
Warning in plot_summary(adm[, .(institution_from_type, patient_id_hashed)]) : 
  Only relevant plotting variables are shown.
 Please inspect your `data` input and specify the variables you would like to plot.
```


`plot_summary(adm[, .(institution_from_type, patient_id_hashed)])`

shijiaSMH, line 1: We need to tweak this warning. As shown, I selected 10 variables, I get the warning is to avoid large data volume, but it's incorrect to show  "No `plot_vars` input provided for large table."

Maybe have another warning for # of selected variables>9?


```

> plot_summary(adm[, .(special_care_unit_death_indicator, total_indirect_cost, province_territory_issuing_health_card_number, readmission, age, gender, admitting_service_mapped, discharging_service_mapped, admit_via_ambulance, patient_service_subservice)])
Warning in plot_summary(adm[, .(special_care_unit_death_indicator, total_indirect_cost,  :
  No `plot_vars` input provided for large table.
 `plot_summary()` will plot all variables (except ID/date-time variables) in `data` input (= 10 variables). This approach is not recommended for large tables with many columns. Instead, we recommend explicitly specifying a subset of variables you want to plot to avoid memory issues/cluttered outputs.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
$`1`

$`2`

attr(,"class")
[1] "list"      "ggarrange"
```

loffleraSMH: This warning is correct. It's shown if users don't explicitly provide a `plot_vars` input. In your example, you provided a subset of adm instead of explicitly specifying `plot_vars = c("special_care_unit_death_indicator", ...)`.

loffleraSMH: Yes, good point - I've added a warning with slightly different wording. This is only relevant when users don't provide a `plot_vars` input:
```
Ignoring any encounter-, patient-, physician-, and hospital-id variables. If you would like to plot them, please explicitly specify those variables using `plot_vars = c(...)"
```

loffleraSMH: I think it's fine if users want to specifically plot ID/date-time variables (for example, it would make sense to do that if someone only has a small number of hospitals and wants to check encounter numbers at each hospital; or they have a date-time variable corresponding to years and want to plot encounters per year). However, I did add a warning in case there are categorical variables with > 50 levels, which also addresses some of your other comments below.

loffleraSMH: See above, this is now implemented as a warning.

shijiaSMH, line 1: Bug: can plot id in derived_variabes table
Also, let's avoid epicare from outputting as well

All codes result in plots
```
plot_summary(data = dv, plot_vars = c("hospital_id"))
plot_summary(data = dv, plot_vars = c("hospital_num"))
plot_summary(data = dv, plot_vars = c("genc_id"))
plot_summary(data = dv, plot_vars = c("epicare"))
```
image

loffleraSMH: This is related to the previous comments: If you explicitly provide `plot_vars = "hospital_id"` (or any other variables) it will always plot any variables the user specifies. This is the intended behaviour because there might be scenarios where it makes sense to plot those variables.
For `epicare`: Good point, I've added that to the list of automatically removed variables if no `plot_vars` input is specified.

loffleraSMH: I've added the following clarification to the vignette:
"Note: If you don't explicitly provide a `plot_vars` input, all encounter/patient/physician/hospital IDs and date-time variables will always be removed automatically. However, if you would like to plot those variables (e.g., to check the number of encounters per hospital), you can explicitly provide them using `plot_vars = "hospital_num"` to avoid those variables from being ignored.

Does this clarify things?

shijiaSMH, line 1: Is it ok to see those 2 warnings when I try to plot all variables?
I think it's a bit repetitive, but alright to see them

```
> plot_summary(data = adm)
Warning in plot_summary(data = adm) :
  Plotting all variables, except encounter/patient/physician/hospital IDs and date-time variables.
 If you would like to plot them, please explicitly specify those variables using `plot_vars = c("variable_name")`

Warning in plot_summary(data = adm) :
  No `plot_vars` input provided for large table.
 `plot_summary()` will plot all variables (except ID/date-time variables) in `data` input (= 33 variables). This approach is not recommended for large tables with many columns. Instead, we recommend explicitly specifying a subset of variables you want to plot to avoid memory issues/cluttered outputs.
```

shijiaSMH, line 1: Flagging date variable can take >30min to run
Some options: 1) user will stop if think it'll take forever, so we do nothing about this; 2) we add a warning whenever the processing time is >10min; 3) we add a warning for all date variables to remind user it can take a while

`plot_summary(data = adm, plot_vars = c("discharge_date_time", "number_of_alc_days"))`

shijiaSMH: saw warning, make sense, closing

shijiaSMH: saw warning, resolve

shijiaSMH: saw it, resolve

shijiaSMH: yes, thx! Pass

shijiaSMH, line 1: this incorrect warning still persists
```

> plot_summary(adm[, .(special_care_unit_death_indicator, total_indirect_cost, province_territory_issuing_health_card_number, readmission, age, gender, admitting_service_mapped, discharging_service_mapped, admit_via_ambulance, patient_service_subservice)])
Warning in plot_summary(adm[, .(special_care_unit_death_indicator, total_indirect_cost,  :
  No `plot_vars` input provided for large table.
 `plot_summary()` will plot all variables (except ID/date-time variables) in `data` input (= 10 variables). This approach is not recommended for large tables with many columns. Instead, we recommend explicitly specifying a subset of variables you want to plot to avoid memory issues/cluttered outputs.

`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
`stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
$`1`

$`2`
```

loffleraSMH: Thanks @shijiaSMH ! Is this better? 

Screenshot 2024-05-17 112516

loffleraSMH: There is already this warning, I think that's sufficient and it's also informative about the source of the issue (large number of categories, rather than something generic about processing time/something too specific for date-times).

Screenshot 2024-05-17 112845

loffleraSMH: @shijiaSMH: The warning is correct. It should always be shown unless you provide an *explicit* `plot_vars` input like this: `plot_summary(adm, plot_vars = c("special_care_unit_death_indicator", ...))`

In your code, you are simply subsetting columns in adm so you are in effect just providing a new (subsetted) table, instead of an explicit input for `plot_vars`.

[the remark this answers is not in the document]
loffleraSMH: @shijiaSMH: Why wouldn't we want to allow users to plot `hospital_id` if they choose to do so? Again, the difference in these approaches is that approach 2 provides an explicit user-provided input for `plot_vars`. I don't want to put any restrictions on what users can plot. The only reason I'm excluding those IDs by default when no `plot_vars` input is provided is to avoid plotting variables that are usually not very informative. Having said that, we can discuss whether or not `hospital_id`/`hospital_num` should be plotted by default (because I think this would generally be more informative than patient-/physician-IDs).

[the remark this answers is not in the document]
loffleraSMH: @shijiaSMH ok so I ended up including hospital_num/id in the plots by default because that’s likely the most common scenario that could cause confusion. 
I tried to make the note on implicit variable selection a bit more obvious in the vignette. I also included an example for when someone might want to plot physician IDs and how to achieve that in the note. I also added a sentence about date-times there. I don’t think we need to elaborate more on that – instead, I’m just referring the user to `plot_over_time`, which should be the function of choice for any exploration of time trends anyway.
Not related to this comment but I also added a small separate section on cell suppression in the vignette. 

The updated vignette is saved on the R drive. Let me know if this is better.

vignettes/plotting_data_exploration.Rmd
shijiaSMH, line 1: Suggest to shorten: This can provide important insights into differences across sites and temporal trends. We recommend generating these plots after cohort creation ~~in order~~ to check for potential data availability/quality issues, outliers, or biases ~~that may have been~~ potentially introduced during cohort inclusion/exclusion steps.

R/plot_over_time.R
shijiaSMH, line 1: Expected to see plot for ID in plot_over_time?
```

> m1 <- merge(adm, phs, by=c('genc_id', 'hospital_id', 'hospital_num'), all.x=T)
> plot_over_time(
+   data = m1, 
+   plot_var = 'mrp_cpso_mapped'
+ )
```

vignettes/plotting_data_exploration.Rmd
loffleraSMH: I removed "in order" as you suggested. I want to avoid repeating "potential/potentially" in the last part, so I shortened it to "biases that may have been introduced during cohort generation" instead.

FYI: I also just made some other changes to the vignette for `plot_over_time` and saved the new version on R.

R/plot_over_time.R
shijiaSMH, line 1: it's a weird case, but should we add a warning when selected plot_var overlaps with out designed subgroups, eg. hospital?

```
plot_over_time(
  data = m1, 
  plot_var = 'hospital_num'
)
```

loffleraSMH: This function does not have any restrictions on what users can plot. `mrp_cpso_mapped` is a character, so the function will automatically plot the % of all `mrp_cpso_mapped` entries that match the last entry (when sorted), which happens to be “na” in this case (or hospital_num = 131 in your example below). So the function plots exactly what’s expected based on the user inputs (it’s just that it doesn’t make sense to plot these things).
There are lots of things that wouldn't make sense to plot, but I'm hesitant to build any assumptions about that into the function (because it would be an endless list). I would view this like any other ggplot2 function: Users can plot whatever they want, and they should be the judge of what makes sense/doesn’t make sense for what they are trying to do. Similarly, users could in principle include any variables in table1 (but a lot of variables wouldn’t make sense to include).

loffleraSMH: Hm, yeah I think in most of these weird cases, users would notice what's wrong based on the output of the function, but just in case, I added this brief warning in case `plot_vars` is identical to any of `time_var`, `line_group`, `color_group`, or `facet_group`.
Screenshot 2024-05-21 094121

R/plot_summary.R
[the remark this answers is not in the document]
loffleraSMH: @shijiaSMH: Good idea! I've added this option when variables are provided as a list input. Note that this only makes sense for categorical variables. I also added an example in the vignette and roxygen.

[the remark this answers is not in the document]
shijiaSMH: Saw those text, certainly better!

vignettes/plotting_data_exploration.Rmd
shijiaSMH, line 1: For 4.2.4 color_group, do we want to remove NA/null/''/' ' when appear in color_group?
image

shijiaSMH, line 1: optional: add a warning when large number of category in color_group?
image

loffleraSMH: I think we should keep missing values in the grouping variables so it's obvious to the users that there are missing values? It's easy enough for the user to remove these from the data input and re-run the function once they notice that there are NA.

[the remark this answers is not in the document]
loffleraSMH: I think the error that's shown by R is already sufficient for users to understand what the issue is? It's just really hard to know what threshold to use as "large" here (because it depends on user's environments...). I included more warnings in `plot_summary` because it comes with a default version that could easily result in issues like this, but I don't think that's the case with `plot_over_time` (the user is actively specifying the variables they want to plot and there isn't really an obvious limit for how many categories make sense/don't make sense).

loffleraSMH: Same as the other comments: I think it's up to the user's judgment how many categories are appropriate for what they are trying to plot? It's obvious from the output that this may not be a desirable plot, so not sure an additional warning is needed.

shijiaSMH, line 1: optional (tiny detail): for this sentence at the end of 4.2.4, can make 'colors' the same format as 'ylimits'. I sense you are referring to function argument here. 

'You can also provide any custom colors and specify ylimits.'

shijiaSMH, line 1: Optional: change text in Additional inputs for facet plots. I initially found it confusing regarding association bwn ylimits and scales inputs.

- Users can provide additional input arguments that are passed to lemon::facet_rep_wrap() (wrapper for ggplot2::facet_wrap()). 
- nrow controls number of rows displayed
- ncol controls number of columns displayed
- scales: 'fixed' (default) or 'free', recommend to specify limits of y-scale when scales='fixed'
- “Free” scales are useful to illustrate within-hospital trends, regardless of between-hospital differences. 
- For example, we may want to visualize the effect of “season” on the encounter numbers (regardless of total number of encounters at each hospital). Here, we plot this in a 2 x 6 facet plot with free y scales:

shijiaSMH, line 1: add a warning when line_group and color_group are different?
Seems color_group overwrites line_group?
```

plot_over_time(
  ipadmdad,
  plot_var = "age",
  line_group = "discharge_disposition",
  color_group = "gender",
  facet_group = "hospital_type",
  min_n = 6,
  show_overall = FALSE
)
```

image

shijiaSMH, line 1: Are you sure this is true? Gender=O is plotted by default but it's small cell?
'Users can additionally provide plot_cat to indicate which category/factor level to plot (by default, the function plots the % in the highest category level).'

> plot_over_time(
+     data = ipadmdad,
+     plot_var = "gender",
+    
+     #func = "%" # not required in this case because gender is a character variable, so the function infers that func should be "%"
+ )

image

shijiaSMH, line 1: can't run this

```
ggplotly(plot_over_time(data = ipadmdad, plot_var = "age"))
Error in match.fun(FUN) : 
  'ggfun("scales_transform_df")' is not a function, character or symbol
```

shijiaSMH: also can't run this

```
> # create plot
> my_plot <- plot_over_time(data = ipadmdad, plot_var = "age")
> # adjust space for x-axis title
> my_plot <- my_plot + labs(x = "\nDischarge Month") 
> # apply additional formatting using layout()
> ggplotly(my_plot) %>%  
+   layout(legend = list(x = 1, y = 0.5)) # change legend position
Error in match.fun(FUN) : 
  'ggfun("scales_transform_df")' is not a function, character or symbol
```

shijiaSMH, line 1: 4.2.8 return_data: It got confusing starting by 'This will return a list of 2 data.table objects...'

I still have a hard time understanding the main difference between res[[1]] vs res[[2]]. I sense you mean res[[1]] is for more granular grouping, whereas res[[2]] is for aggregated grouping. 

Another pt, do you think the results below are as expected? Can you pls explain the logic of grouping if possible?
Ex.  I don't understand why in ur ex, hospital_type is shown, but in my ex, hosptial_num is shown
```

res2 <- plot_over_time(
  ipadmdad,
  line_group = 'discharge_disposition',
  color_group = "gender",
  func = "median",
  plot_var = "age",
  return_data = TRUE
)
> res2[[1]]
            month discharge_disposition gender hospital_num median_age  n
    1: 2018-04-01                     4      F            1       79.0 11
    2: 2018-04-01                     4      F            2       79.0  5
    3: 2018-04-01                     4      F            3       80.0 10
    4: 2018-04-01                     4      F            4       80.5  6
    5: 2018-04-01                     4      F            5       75.5  8
   ---                                                                   
14996: 2023-03-01                    90      M            3         NA NA
14997: 2023-03-01                    90      M            5         NA NA
14998: 2023-03-01                    90      M            8         NA NA
14999: 2023-03-01                    90      M           12         NA NA
15000: 2023-03-01                    90      O            5         NA NA
> res2[[2]]
           month gender hospital_num median_age  n
   1: 2018-04-01      F            1       75.0 31
   2: 2018-04-01      F            2       77.0 39
   3: 2018-04-01      F            3       83.0 25
   4: 2018-04-01      F            4       80.5 32
   5: 2018-04-01      F            5       76.0 39
  ---                                             
1736: 2023-03-01      O            1         NA NA
1737: 2023-03-01      O            5         NA NA
1738: 2023-03-01      O            6         NA NA
1739: 2023-03-01      O            8         NA NA
1740: 2023-03-01      O           10         NA NA
```

loffleraSMH: Thanks, it's fixed now.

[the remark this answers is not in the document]
loffleraSMH: Same as above: For any grouping variable, I think it's better to keep NA, so at least users can see in the output that there are NAs (and there might be cases where they actually want to plot groups with NA; if not, they can easily remove them before plotting). 
For small cell counts: I previously considered setting `min_n` to 6 by default but ended up deciding against it because a) when plotting counts, it's actually really useful to see cells with low counts (e.g., n = 0 reflects data availability issue), and b) more generally, since these plots are meant for data exploration, I want to avoid the function from automatically removing things that can actually be informative (like in your case, it can be useful to see that the categories "O"/"U" exist for gender, users can then make their own judgment about what should be included/excluded). 

I did however make a different change that's somewhat related: In the plot you shared, some of the lines don't show up (e.g., for gender = "U") because there is only a single data point in that plot, so `geom_line()` doesn't show anything because there are no points that could be connected with a line. I just pushed a fix so the function will now plot any isolated data points like that with `geom_point()`. I think this is the most transparent way of dealing with those cases? I also added a warning in cases where this happens. Let me know what you think. 

Screenshot 2024-05-23 180042

Screenshot 2024-05-23 175520

loffleraSMH: I changed this section to bullet points, but with some small edits to your suggestion (e.g., I combined `nrow` and `ncol` into a single bullet point, and I don't think using "fixed" scales means we necessarily recommend specifying the limits of the y-scale).

loffleraSMH: Right, so the problem here isn't that `line_group` and `color_group` are different. They can be different, for example, see this plot: 

Screenshot 2024-05-23 183056

However, `line_group` needs to be nested within `color_group` (since color coding is applied to individual lines, not within line, that's generally true for all `ggplots`). In your example: You are plotting separate lines by discharge disposition but then you are trying to apply color by gender (so you are trying to apply color within lines). I've added this warning when the nesting structure of the specified grouping variables is implausible:

Screenshot 2024-05-23 185701

loffleraSMH: Yes, sorry wording is confusing here: "Highest category" as in when you sort the variable levels themselves (not by frequency), the last level will be plotted ("O" in this case because it's sorted alphabetically, not by frequency). The reason this generally makes sense is because it will by default plot % TRUE for any logical variables. I've changed the wording in the vignette to "by default, the function plots the % in the last category level, e.g., `TRUE` for logical variables". Is that better?

loffleraSMH: Looks like this might be an issue with `ggplot2` / `plotly` versions? https://github.com/tidyverse/ggplot2/issues/5285
Does `ggplotly` work for other ggplots for you right now? If not, this probably isn't caused by anything in the `Rgemini` plotting functions, so let's try to fix this independently of the current issue.

loffleraSMH: Correct, the 1st item is the grouping by all grouping variables, whereas the 2nd one corresponds to what would be shown as the "overall" line in the plots. So the output depends on the grouping variables you specify (you didn't include `hospital_type`, so that's why it doesn't exist in your output). I've updated the vignette to make this more explicit (and also fixed an error in the wording). Updated vignette is on the R drive. Is this better? 
FYI, I also made a small change to the output variable name "month" (which is now shown as `discharge_month` to be more specific).

loffleraSMH:
@shijiaSMH: I've implemented one additional enhancement for `plot_summary` to allow users to specify a `facet_group` when they only plot a single `plot_var`. This is useful to check distributions by individual sites (or other grouping variables). Could you take a quick look at section 3.3. in the vignette to see if that makes sense? Thanks!

vignettes/plotting_data_exploration.Rmd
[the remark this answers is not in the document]
shijiaSMH: saw error, resolve

vignettes/plotting_theme.Rmd
```diff
@@ -0,0 +1,309 @@
...
+# generate barplot
+my_plot <- ggplot(data = my_data, aes(x = group, y = mean, fill = group)) +
+  geom_bar(stat = "identity", fill = gemini_colors(1)[1:3]) +
+  scale_y_continuous(name = "Mean", expand = c(0, 0))
+
+# original ggplot theme
+my_plot + ggtitle("Original plot")
+```
+
+```{r}
```
youx-unity, line 79: Hi, I would suggest here to add `library(ggthemes)`otherwise it would report error as `could not find function "theme_foundation"`

loffleraSMH: Hi @youx-unity, note that calling `library(ggthemes)` is not required when the `Rgemini` package is installed (because `ggthemes` is already imported by the function `plot_theme()`). You can test this by installing `Rgemini` from this branch (or alternatively, you can load the required packages that are listed on lines 35-40 as a temporary solution but we won't need to add this here).

vignettes/plotting_theme.Rmd
```diff
@@ -0,0 +1,309 @@
...
+- viridis color maps: https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html
+- color brewer: https://colorbrewer2.org/#type=sequential&scheme=BuGn&n=3
+
+
+<!-- remove border around images -->
+<style>
+img {
+border: 0;
+}
+</style>
```
youx-unity, line 309: Hi Anne, I reviewed the knitted vignette from here: R:\GEMINI\Research Projects\Rgemini\. I did some testing with plot_theme() and tried all the GEMINI colours. The Vignette is very well documented and I liked the layout. Everything has been listed in bullet points and easy to find information. The example codes are very informative. I ran the codes and there are no errors. All the GEMINI colours not only can be referenced by colour number but also by colour names, such as "Lavender Lagoon", which I really like. All the names are beautiful and elegant.  

The only issue I had was the `library(ggthemes).` The remote install doesn't work so I used git clone and sourced the function for the testing. Besides that, everything looks great. The codes run smoothly with no errors and all the colours work perfectly.

loffleraSMH: Hi @youx-unity, thanks so much for the review! Great to hear everything worked well for you. Regarding the `library(ggthemes)` issue, this will be solved once everything has been merged into `master` (and then you can install the newest version from there, which should prevent this issue).  

I will wait for @shijiaSMH to finalize her review on the other plotting functions. Once everything has been approved, I'll merge the pull request and close this issue.

Thanks again to both of you for taking on these reviews! :-)

PR #119, opened 2024-05-15
85 fix error when input has labelled class

### LINKED ISSUE #85: [BUG]: max_pairwise_smd does not work with labelled column
## New Feature Request

When using `table1()` to generate tables, it can be useful to label column names for better display in the output html table. e.g.

```
# convert gender to binary sex
cohort$sexf <- cohort$gender=="F"
# improve readability in output table
label(cohort$sexf) <- "Sex Female"
```
Currently the `max_pairwise_smd` function will give error when the input column is labelled:
```
Error in (vartype == "numeric") || is.integer(x$value) : 
  'length = 2' in coercion to 'logical(1)'
```
Because it assumes the input column has only one class while labelled column can have two:
```
> class(cohort$sexf)
[1] "labelled" "logical" 
```

This can be improved by something like:
```
  fn <- if (("numeric" %in% vartype) || is.integer(x$value)) {
    stddiff.numeric
  } else if ("logical" %in% vartype) {
    stddiff.binary
  } else if (any(c("factor", "character") %in% vartype)) {
    stddiff.category
  }
```

guoyi-smh:
This PR closes #85 

After further investigation, the error was due to `label()` in `table1()` masked by `Hmisc::label()`, you can find the details in [this comment](https://github.com/GEMINI-Medicine/Rgemini/issues/85#issuecomment-2113353757). Since `Hmisc` is a widely used package for EDA, in case others will encounter similar issue.

### DIFF
--- a/R/cell_suppression.R
+++ b/R/cell_suppression.R
@@ -42,6 +42,7 @@ max_pairwise_smd <- function(x, name, round_to = 3, ...) {
   pairs <- unique(x$L1) %>% combn(2, simplify = FALSE)
 
   vartype <- class(x$value)
+  vartype <- vartype[!vartype=="labelled"]
 
   fn <- if ((vartype == "numeric") || is.integer(x$value)) {
     stddiff.numeric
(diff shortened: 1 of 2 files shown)


wankiewiczm-smh:
Used the updated code and found no issues when creating a `table1`. I also loaded the `Hmisc` package and used the incorrect `label` function, `table1` still rendered as expected.

Everything works fine on my end, great work @guoyi-smh!

guoyi-smh:
Thank you Matthew for reviewing and testing!

PR #121, opened 2024-05-17
71 icu entry as outcome

### LINKED ISSUE #71: Add option `icu_entry`  to return ICU admission as a clinical outcome
## New Feature Request

Currently, `icu_entry` returns whether or not an encounter had been admitted to ICU, including those with ICU_admit time <= IP_admit time. 
In research projects, ICU admissions are often a study outcome, which should not take place at the same time as the exposures of interest. Therefore, a new feature will be added to the `icu_entry` function to enable returning ICU admissions post exposure.
- The new feature should allow the options to return ICU-admission as a descriptive feature (current version) or ICU-admission as a clinical outcome.
- Ideally, it should also allow users to specify a buffer window in which ICU-admissions will not be considered as a TRUE ICU admission (thus coded as icu_entry=F) when it is a clinical outcome.

gemini-wenb:
Closes #71 
This PR implemented a new feature to `icu_entry()`. The new feature provides an option to compute ICU admission as a clinical outcome. 
In research projects, ICU admissions are often a study outcome, which should not take place at the same time as the exposures of interest. With this new implementation, `icu_entry()` has the options to return ICU-admission as a descriptive feature (i.e. not excluding entries prior to or at IP admission) or ICU-admission as a clinical outcome.

Key changes:
1. added option to return ICU as a clinical outcome or not (param `as_outcome` )
2. added option to customize a different cutoff time (default is IP admission time) to exclude records from being considered in the identification of ICU entries (param `exclude_cutoff`)
3. updated parameter naming from `window` to `entry_since_cutoff` to avoid potential confusions as a result of the new features
4. updated documentation with examples for common use cases
5. updated function warning messages on invalid date-time and exclusion details
6. created unit tests for this function: `test-icu_entry.R` (Reviewer, please feel free to suggest additional unit tests if there is any important testing case that has not been covered here).


marchenaromk:
Hi Bijun, thanks for including the test cases! They were very helpful and all test cases passed. Just one quick thought: should ICU entries that occur prior to admission be considered `NA `rather than `False` when `as_outcome = T`? This would make a clear distinction between entries that truly were not admitted to the ICU vs entries that are removed due to a specified criteria

marchenaromk:
Thanks that makes a lot of sense! Especially in the case where there could be multiple ICU entries. Thanks for the clarification! Nice work!

marchenaromk (approved):
Test cases pass and I can't think of another test case that would be relevant for this function.

gemini-wenb:
> Test cases pass and I can't think of another test case that would be relevant for this function.

Thank you for reviewing the function, Kayley! I will merge this :)

PR #126, opened 2024-05-23
72 update comorbidity function reference

### LINKED ISSUE #72: Investigate elixhauser imputation
Currently when there are no diagnoses at admission, `compute_charlson_comorbidity` imputes 0 for these encounters.

Elixhauser comorbidity adopts the same approach although the lowest possible Elixhauser score is -20. However this distribution matches what we expect from the derivation paper:

image

If this is the correct imputation strategy, note this in the function documentation. Otherwise adjust the imputation strategy as needed.

guoyi-smh:
closes #72 

Fixed references in `charlson_comorbidity_index` and `elixhauser_comorbidity_index`.

### DIFF
--- a/man/elixhauser_comorbidity_index.Rd
+++ b/man/elixhauser_comorbidity_index.Rd
@@ -36,7 +36,8 @@ By default, for each encounter, outputs the comorbidity score. If \code{raw_como
 outputs a wide \code{data.table} with a column for each comorbidity for each encounter.
 }
 \description{
-Based on the methodology from \href{https://pubmed.ncbi.nlm.nih.gov/19433995/}{Elixhauser A et al. 1998}
+Based on the methodology from \href{https://www.jstor.org/stable/3766985}{Elixhauser A et al. 1998},
+\href{https://www.jstor.org/stable/3768193}{Quan H et al. 2005}, and \href{https://doi.org/10.1097/MLR.0b013e31819432e5}{van Walraven C et al. 2009}
 using ICD-10 codes as the mapping algorithm,
 implemented in the R \href{https://ellessenne.github.io/comorbidity/index.html}{comorbidity} package
 by \href{https://doi.org/10.21105/joss.00648}{Gasparini, 2018}.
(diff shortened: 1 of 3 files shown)


man/elixhauser_comorbidity_index.Rd
```diff
@@ -52,7 +53,10 @@ This affects comorbidities present in a patient with different degrees of severi
 }
 \references{
 \itemize{
-\item{Elixhauser A, et al. Med Care, 1998. https://doi.org/10.1097/MLR.0b013e31819432e5}
+\item{Elixhauser A, et al. Med Care, 1998. https://www.jstor.org/stable/3766985}
+\item{Quen H, et al. Med Care, 2005. https://www.jstor.org/stable/3768193}
+\item{Van Walraven C, et al. Med Care, 2009. https://doi.org/10.1097/MLR.0b013e31819432e5}
```
youx-unity, line 58: Hi Yishan, I saw Vaakesan commented this paper in the description. But I don't see other paper reference in the code review description. From my understanding, the request is to update the reference from 1998 paper to 2009 paper from Van Walraven for Elixhauser  method? Please correct me if I'm wrong. I didn't find any instructions about charlson score paper references. Are those being communicated offline?

guoyi-smh: Hi Shirly, Thank you for your review! 
The methods in the functions (both Charlson and Elixhauser) are built on top of the three papers here. The raw comorbidities were developed in 1998 paper, the mapping from ICD10 to the Elixhauser comorbidities were developed in 2005 paper and the weights to compute a single integer score was derived in 2009 paper. I think it would be better to list them all. You can find more details in [this round table presentation slides deck](https://docs.google.com/presentation/d/1bux3tDlk3vSYsBbF-TkeYOZoOey4aS9Q/edit#slide=id.p1)

man/elixhauser_comorbidity_index.Rd
youx-unity, line 1: Hi Yishan, Thank you for the clarification. This looks great to me.

PR #151, opened 2024-07-26
DRAFT: 139 data coverage

loffleraSMH:
This is a first version of the data coverage function (draft PR, please don't merge yet).

The following things will need to be fixed:
- Currently, this function only works if `hospital_id` exists in the DB (i.e., it wouldn't work for HPC users); once the availability lookup table has been updated in `report_db`, I'll check the function to make sure it works for all users.
- The location of the data availability lookup table will also change in the DB, so I will need to fix the query accordingly.

For now, please review the function to make sure everything works as expected and the function documentation is clear.

You can use the following code as a starting point for testing:

```
library(DBI)
library(RPostgreSQL)
library(getPass)
library(dplyr)
library(data.table)
library(Rgemini)
library(lubridate)
library(ggplot2)

drv <- dbDriver("PostgreSQL")
db <-DBI::dbConnect(drv,
                    dbname = "report_db_v3",
                    host = "prime.smh.gemini-hpc.ca",
                    port = 5432,
                    user = getPass("User:"),
                    password = getPass("Enter Password:"))

cohort <- dbGetQuery(db, "SELECT * FROM admdad WHERE discharge_date_time >= '2018-04-01 00:00';") %>% data.table()

source("~/GitHub/Rgemini/R/data_coverage.R") # adjust path according to your Rgemini folder
source("~/GitHub/Rgemini/R/utils.R") # adjust path according to your Rgemini folder

## Default function (create all plots)
coverage <- data_coverage(db, cohort, c("admdad", "transfusion"))

## Suppress plots (will only return single table with encounter-level flag
coverage <- data_coverage(db, cohort, c("admdad", "radiology", "ipscu"), plot_timeline = FALSE, plot_coverage = FALSE))

```


R/data_coverage.R
```diff
@@ -0,0 +1,574 @@
...
+#' }
+#'
+#' @param dbcon (`DBIConnection`)\cr
+#' A database connection to any GEMINI database.
+#'
+#' @param cohort (`data.frame` or `data.table`)
+#' Cohort table with all relevant encounters of interest, where each row
+#' corresponds to a single encounter. Must contain the following columns:
+#' - `genc_id`: GEMINI Encounter ID
+#' - `hospital_num` | `hosital_id`: Hospital identifier
```
danikanou, line 52: *Super* minor, hosital_id instead of hospital_id

loffleraSMH: Good catch! I just pushed a fix.

PR #157, opened 2024-10-01
77 cohort creation

### LINKED ISSUE #77: Cascading Frequency Table By Exclusion Criteria and Hospital
## New Feature Request

Function creating table showing number/percent of encounters/rows at each inclusion/exclusion step during cohort creation. 

Some initial code snippets can be found below. Ideally, we can add the following optional input arguments:
-  user-defined stratification variable (`character`) corresponding to separate columns in output table (typically, `hospital_id` or `hospital_num`, but could also be `hospital_type` etc. so let's keep it flexible)
- `show_prct` flag allowing users to control whether to include percentages (default = TRUE), otherwise, will only show counts
- [as discussed with Yishan, leave this for now to keep function logic simple: `idx` (`numeric` vector) that can be used to indicate if multiple steps are part of the same inclusion/exclusion step (i.e., they should have the same denominator), e.g., inclusion 2a and 2b for COVID vs. flu patients (both should be included but numbers should be listed separately, with percentages being based on cohort from inclusion step 1 in both cases; e.g. `idx = c(1, 2, 2)`.] 

```
library(dplyr)
library(data.table)

set.seed(1)
my_data <- Rgemini::dummy_ipadmdad(100) %>% data.table()


cohort_creation_table <- function(cohort, stepnames = NULL, stepnames_short = NULL) {
  
  # get exclusion steps (will be shown as -n (-X%) to show what number/% of patients was excluded)
  excl <- grepl('^ex', stepnames_short, ignore.case = TRUE) 
  
  steps <- sapply(cohort, nrow)
  
  # create table
  tabs <- data.table(steps, lag(steps))
  tabs[, percent.change := round(1000*steps/V2)/10] 
  
  # for exclusion steps, show removal as -n (-X%) 
  if (sum(excl) > 0) {
    tabs[seq(sum(excl==F)+1, nrow(tabs)), percent.change := -1*round(1000*(1-steps/V2))/10] 
    tabs[seq(sum(excl==F)+1, nrow(tabs)), steps := -(V2-steps)] 
  }
  tabs[is.na(tabs)] <- ""
  tabs[, ':='(`N (%)` = ifelse(!is.na(percent.change), 
                               paste0(steps, " (", percent.change, "%)"), 
                               steps),
              steps = NULL, percent.change = NULL, V2 = NULL)]
  
  
  tabs <- cbind(stepnames, tabs)
  
  
  # add row with final cohort number (only needed if previous step was showing
  # exclusion
  if (excl[length(excl)] == TRUE) {
    tabs <- rbind(tabs, data.table(stepnames = "Final cohort",
                                   `N (%)` = nrow(cohort[[length(cohort)]])))
    stepnames_short <- c(stepnames_short, " ")
  }
  
  tabs <- cbind(stepnames_short, tabs)
  
  colnames(tabs) <- c("", "Cohort creation step", "N (%)")
  
  return(tabs)
}

cohort_table <- cohort_creation_table(
  cohort = list(my_data, 
                my_data[gender == "F"], 
                my_data[gender == "F" & age >= 80]),
  stepnames = c("All GEMINI encounters", 
                "Gender = Female", 
                "Age < 80"),
  stepnames_short = c("Incl. 1", "Incl. 2", "Excl. 1")
)

print(cohort_table)
```

Image

loffleraSMH:
closes #77 

I added a new function `cohort_creation()` that creates a cohort data.table based on inclusion/exclusion criteria and also returns a table showing number (%) of entries that were included/excluded at each step.

Please check the function documentation for clarity and make sure that the inputs/outputs are intuitive and user-friendly.
 
You can use this code as a starting point for testing the function:

```
library(dplyr)
library(data.table)
library(Rgemini)

source("H:/GitHub/Rgemini/R/cohort_creation.R") # pull from branch 77
my_data <- Rgemini::dummy_ipadmdad(10000, n_hospitals = 5) %>% data.table()

cohort <- cohort_creation(
  cohort = list(my_data, 
                my_data[gender == "F"], 
                my_data[age > 65],
                my_data[grepl("^7", discharge_disposition)]),
  labels = c("All GEMINI encounters", 
             "Gender = Female", 
             "Age > 65",
             "In-hospital death"),
  exclusion_flag = c(FALSE, FALSE, FALSE, TRUE)#,
  #group_var = "hospital_num" # optional: stratify by hospital
)

cohort_data <- cohort[[1]] # data.table with final cohort
print(cohort[[2]]) # inclusion/exclusion table 
```


R/cohort_creation.R
```diff
@@ -0,0 +1,218 @@
...
+#' corresponding entries will be removed and the number (%) of rows that were
+#' removed (rather than included) will be shown.
+#'
+#' By default, all cohort steps will be interpreted as inclusion steps.
+#'
+#' @param show_prct (`logical`)
+#' Flag indicating whether to show percentage values. If `FALSE`, only raw
+#' counts will be shown. Note that the percentages always reflect the % change
+#' relative to the N in the *previous* inclusion/exclusion step.
+#'
```
marchenaromk, line 38: Do you want to add a note here that the default for `show_prct` is `TRUE`

marchenaromk:
1. Tested `group_var` flag. Works as expected 
2. Tested non-present column, gave warning as expected and did not produce table
3. There is no `final cohort` row if `Inclusion` is final step. Should we indicate a best practice in the description or add `Final Cohort` line in general?
4. Function fails if `my_data` table is a `data.frame`
5. Extreme case, but I'm wondering if there would ever be a scenario where N < 6 for a single hospital?

marchenaromk:
1. I think leaving it as is with final cohort only included for exclusion last step makes sense! Thanks for the clarification.
2. My mistake! I subset the data.frame as it should be, and it works.
3. I tested the cell suppression functionality and it works well! Tried where N = 6 and N < 6 and cell suppression works as expected.

loffleraSMH:
Ok great, thanks so much again for the review @marchenaromk! Let me know if your are still testing anything else, or if you are done with the review and I can go ahead and merge this.

loffleraSMH:
Oh also, @marchenaromk: I just added some [unit tests](https://github.com/GEMINI-Medicine/Rgemini/blob/77_cohort_creation/tests/testthat/test-cohort_creation.R) for this function as well. You don't need to run them (the CI/CD workflow automatically checks the tests to make sure they all pass) but you can take a quick look to see if I missed any other tests that would be worth adding.

loffleraSMH:
> I'm finished reviewing! I can't think of any other scenarios we should be testing. Great work with this function! It's so useful!

Ok awesome, glad you find it useful :-) Perfect, I'll go ahead and merge. Thanks again for the review!

PR #173, opened 2025-01-29
DRAFT: 142 stats vignette

loffleraSMH:
This is a draft PR to facilitate a first round of review for the vignette.


vignettes/stats_intro.Rmd
```diff
@@ -0,0 +1,1210 @@
...
+cohort[is.na(mlaps), mlaps := 0]
+```
+
+
+**Note that it's important to distinguish between "data coverage" (overall data volume for certain tables) vs. "missing values" (individual entries with `NA` values for certain table columns).** Even when data coverage is generally high, individual entries may still have missing values. For time periods with high data coverage, we can sometimes make certain assumptions about missing values based on the clinical context, allowing us to perform single-value imputation (e.g., imputing all missing `mlaps` values with 0). There are many cases where we cannot make these assumptions. Make sure to discuss how you handle missing values with your PI and Scientific Advisor. 
+
+<details>
+<summary>***Expand this section to learn more about different types of missingness.***</summary>
+
+In our scenario, missingness of `mlaps` represents an example of systematic missingness - also referred to as **"missing not at random" (MNAR)**. This means that the missingness is directly related to other predictor variables (e.g., patients' illness severity) and/or outcomes (e.g., in-hospital death).
```
surainSMH, line 546: I don't think we should introduce the concept of missingness through mLAPS imputation of 0. I agree that technically we believe mLAPS are MNAR because they are missing for a reason. However, we don't impute mLAPS with 0 not *because* it's systematically MNAR, but because we have a theoretical justification to do so. MNAR data are generally less reasonable to impute than MAR and MCAR. I think we need a separate vignette discussing missingness (MCAR, MAR, MNAR), multiple imputation, conditional mean imputation,, missing indicator method, etc.

vignettes/stats_intro.Rmd
```diff
@@ -0,0 +1,1210 @@
...
+
+```{r}
+data[, incl3_pneumonia := grepl("^J12|^J13|^J14|^J15|^J16|^J17|^J18", ipdiagnosis_mrdx)]
+```
+
+<details>
+<summary>**Expand this section for additional considerations when filtering by ICD-10-CA codes**</summary>
+
+**1) What diagnosis codes to include/exclude?** 
+This is usually determined by an SME who will provide you with a list of relevant ICD-10-CA codes. In our example, all ICD-10-CA codes from J10-J18 should be included for pneumonia. Usually, you should filter for all diagnosis codes that **start** with the same characters (i.e., including children codes like "J12.8"), unless otherwise specified. If you are unsure, please clarify with the SME.
```
surainSMH, line 288: Would also make reference to CCSR here. It is something that pops up in many projects and adhocs.

vignettes/stats_intro.Rmd
```diff
@@ -0,0 +1,1210 @@
...
+**1) Data coverage:** Certain hospitals may not have any (administrative) data during certain time periods.
+
+**2) Patient volume:** Some hospitals might have a very large/small number of encounters overall, which is important to consider for the analyses below (e.g., small community hospitals will generally contribute fewer encounters to the overall cohort than large academic hospitals).
+
+**3) Changes in the cohort over time:** For example, you might observe sudden drops/increases in encounter numbers, which might be due to seasonal changes, COVID-related factors, changes in the GEMINI cohort definition (e.g., expansion from GIM to all-medicine), or potential biases introduced during cohort inclusion/exclusion steps (see below).
+
+</details>
+
+
+In our example data, we can see that sites 6, 9, 10, and 12 do not have any encounters during certain time periods, indicating gaps in data coverage. We will therefore exclude these sites to make sure we have consistent data timelines across all hospitals.
```
surainSMH, line 228: I think we need to explain how to think about excluding sites with incomplete data coverage. It's unclear to me that complete coverage by hospital over the time period is necessary to answer this question. Do we expect this to induce a bias with respect to age -> death relationship? Why did we not just limit to time periods with those data available? 

The decision to exclude good data because of other missing time periods is specific to the research question.

vignettes/stats_intro.Rmd
```diff
@@ -0,0 +1,1210 @@
...
+
+Confounders are not the only variables that can affect the relationship between age and in-hospital mortality or LOS. Mediators, or intermediate variables, lie in the causal pathway between the exposure variable and outcome. In this analysis, a mediator would be caused by age and is a cause of the outcome variables. This type of analysis, mediation analysis, is beyond the scope of this study, but it is important to understand the distinction between controlling for a confounder that could bias results, and including mediators that block the path to an outcome: [MacKinnon D. Res Soc Work Pract, 2011](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3366634/)
+
+
+## Cohort definition
+
+Based on the conceptual model discussed above, we will define our cohort as follows:
+
++ **Inclusion steps**
+  - Incl. 1: All GIM encounters discharged between Apr 2018 - Apr 2020
```
surainSMH, line 145: Would briefly describe GIM: our data are dept of medicine and ICU. GIM is the biggest medicine department. It treats mainly complex multimorbid patients. 40% of emerg admissions end up in GIM. It is very heterogeneous (unlike other services, which specialize in 1 thing like cardiology or nephrology). We often look at GIM because that's what many of our collaborators are interested in. Also has nice characteristics beyond the scope of this document that allow for physician-level inference (quasirandomization), so it's a good cohort to examine quality of care.

vignettes/stats_intro.Rmd
```diff
@@ -0,0 +1,1210 @@
...
+- For more information, check out these resources: [DAGs: A simple introduction with simulations in R](https://arelbundock.com/posts/acmq_en_06_dag/), [Applications of DAGs in Causal Inference](https://www.r-bloggers.com/2018/08/applications-of-dags-in-causal-inference/)
+</details>
+
+## Third Variables
+While it is possible to estimate the effect of age on mortality on its own, often called an unadjusted model, it is common practice to control for variables that could have a direct or indirect cause on the exposure and outcome. These variables are called confounders (also referred to as covariates) and controlling for these variables debiases the association between the variables of interest: [Wysocki A et al. Adv Meth Pract Pyschol Sci, 2022](https://journals.sagepub.com/doi/10.1177/25152459221095823#:~:text=After%20outlining%20a%20causal%20structure,of%20the%20predictor%20and%20outcome). 
+
+  - Gender can affect the age that one acquires pneumonia, along with the severity of illness            experienced when pneumonia is acquired: [Corica B et at. Intern   Emerg Med, 2022](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9294783/).
```
surainSMH, line 128: Echoing my comment in today's meeting, I think this is a tough example to introduce DAGs with. DAGS are most easily understand as cause - effect diagrams. mLAPS and Charlson do not cause Age - time is the only thing that causes age. 

I like how you've put in gender and justified w selection bias - this seems reasonable to me (but let's confirm this is valid). I think we can keep mLAPS and Charlson, but consider them as mediators w arrows from age to each, and each to death. 

Open to feedback about how we update the modeling to match the DAG. Would be good for those to be married.

surainSMH: We can make reference to them as mediators while maintaining the "holding X constant" emphasis on interpretation. We can refer them to another future vignette for bells and whistles to rigorously test the mediation effect etc.

vignettes/stats_intro.Rmd
loffleraSMH: Good catch! I rewrote that section: It now focuses on how we deal with missing `mLAPS` first, and then I added a separate section with **"Other considerations"**, including MAR vs. MNAR with a reference for more details on that.

Upon rereading this, I'm also not sure we really want to recommend imputing missing mLAPS with 0. I think it would be reasonable to just do a complete-case analysis here (although the data coverage checks become a little bit less relevant then...) or recommend multiple imputation (which we can briefly illustrate in the modelling section)?

During Academic vs. Community review process, we compared patients with true mLAPS = 0 vs. those with missing mLAPS within 24hr. Those 2 groups look fairly different (e.g., those with missing mLAPS include a higher % of palliative patients, patients transferred from other institutions, and they also tend to have higher CCI).
mLAPS0_baseline_characteristics. Based on that, it's not really clear our justification for imputing with 0 holds true. And for a question like this, it might be hard to justify why we would recommend single-value imputation given the large sample size  and only 2% missing mLAPS?

@surainSMH, @gemini-wenb, and @marchenaromk let me know what you think and I'm happy to change this section accordingly.

loffleraSMH (commented):
Thanks @surainSMH for the feedback! I've addressed all comments (except DAG, which @marchenaromk is going to update once we've updated the modelling section). I had one follow-up question regarding mLAPS imputation, which might require some discussion (see below).

PR #176, opened 2025-02-04
110 readmission faq

### LINKED ISSUE #73: Improve palliative vs. chemo documentation in readmission vignette
## Improve documentation

The current vignette for the readmission function provides some general rationale for the palliative and chemotherapy flag, but it says that the rationale for the chemo flag is the same as for the palliative flag. 

image (7)

Thus, it is not clear why palliative care is removed from both the numerator and denominator, whereas chemotherapy is only removed from the numerator:

Screenshot 2024-01-31 161139

We should add some additional clarification on why palliative care (but not chemotherapy) is removed from the denominator. Something along those lines (please edit & shorten):

The main distinction is that palliative care would usually result in death, and thus, we wouldn't want to check whether those encounters are readmitted again after an episode of care that involved palliative care. So in that sense, there is some overlap between the palliative and death flag (in neither case, checking for readmission would be fair because the patient by definition can't be readmitted). By contrast, episodes of care involving chemotherapy could in principle result in a subsequent readmission (but note that if that readmission itself involves chemotherapy, it wouldn't be counted as a readmission, otherwise, we'd likely be inflating readmission rates for cancer patients; however, if someone is readmitted for a non-chemotherapy encounter, that would count as a readmission)

### LINKED ISSUE #110: Add FAQ section to readmission vignette that lists reasons for `NA`
## Improve readmission documentation

Please add a brief FAQ section at the end of the readmission vignette with bullet points listing all possible reasons for "why are there `NA` in the derived readmission flags"? The vignette is very long and the different reasons for `NA` are distributed throughout the vignette, so it would be good to have a brief overview of the different reasons for `NA` where users can get a quicker answer to this commonly asked question. 

Here is a list of all reasons:
- Any `genc_ids` during the buffer period (end of data availability period) are returned as `NA`
- Only the last encounter of each episode of care is assigned readmission flag, all previous encounters of the same episode of care have `readmission = NA`
- If encounter is followed by an acute care transfer to another hospital but transfer encounter does not exist in GEMINI data (e.g., transfer to non-GEMINI hospital), all encounters of that episode of care have `readmission = NA` (because we don't know how long patient was hospitalized for during transfer encounter / we can't attribute readmission to the current GEMINI encounter)
- The following flags (if set to `TRUE`) will result in `NA` for some encounters: `death`, `palliative`, `mental`, and `signout`

Ideally, we can link each list item to the relevant section in the vignette so users can easily find more details.

loffleraSMH:
This PR addresses 2 separate issues related to the readmission vignette:

1. I've added a FAQ section to the readmission vignette that provides a quick overview of reasons for `NA` in the readmission flag (closes #110). Please make sure the list of reasons is complete and easy to understand.
2. I have also improved the section on CIHI flags for clarity. Specifically, I provided a clearer rationale for the palliative and chemo flags, which were previously listed as "same rationale" despite differences in removal from denominator vs. numerator (closes #73). I also made some small edits to the other flags to improve the wording. Please make sure the edits are correct.

Finally, I also fixed the reference to the buffer section, since this previously didn't link correctly (needs to be a heading).

### DIFF
--- a/vignettes/epicare_and_readmission.Rmd
+++ b/vignettes/epicare_and_readmission.Rmd
@@ -481,7 +481,8 @@ These steps are meant as a general guideline, but you may need to consider addit
 
 ## Why are there missing values in the readmission flags?
 
-Missing values (`NA`) in the returned readmission flags indicate that the encounter was removed from the denominator and should not be included in any analyses of readmission rates. There are several reasons for `NA` in the derived readmission flags:
+Missing values (`NA`) in the returned readmission flags indicate that readmission status for these encounters cannot be determined. Encounters with `NA` are excluded from the denominator in readmission rate calculations, and thus, should be removed from any readmission analyses. There are several reasons for `NA` in the readmission flags:
+
 
 1) **Buffer period:** Any `genc_ids` that were discharged during the buffer period (end of data availability period for a given hospital) are returned as `readmit = NA` because readmission cannot be reliably determined for these encounters (see [here](#sectionBuffer) for more details). Note that longer readmission windows will lead to a longer buffer period (i.e., more `NA`).
 
--- a/vignettes/epicare_and_readmission.Rmd
+++ b/vignettes/epicare_and_readmission.Rmd
@@ -55,7 +55,7 @@ knitr::include_graphics("figures/epicare_out_2.png", dpi = 100)
 ```
 
 
-## `readmission()`
+## `readmission()` {#sectionIntro}
 
 The function **`readmission()`** computes whether or not a patient associated with an epicare was readmitted to a GEMINI hospital within a specified time window. For each hospitalization (`genc_id`), the function returns its corresponding: `AT_in_occurred`, `AT_out_occurred`, `epicare`, and `readmitX`. `readmitX` is a boolean variable denoting whether readmission took place during a time window X-days post-discharge. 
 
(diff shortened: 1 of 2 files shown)


vignettes/epicare_and_readmission.Rmd
```diff
@@ -485,3 +475,19 @@ There may be more complex scenarios where the inclusion criteria for index admis
 5) Calculate readmission rates: Filter the output of the `readmission` function based on your index admissions (cohort created in step 1) in order to get readmission rates of surgery patients to GIM. This step is important to avoid that any admissions following GIM epicares (e.g., GIM to GIM readmissions or GIM to surgery readmissions) are included in the readmission rates. 
 
 These steps are meant as a general guideline, but you may need to consider additional aspects for a given research project. Please make sure you carefully consider the implications of the decisions you make when building your `restricted_cohort`. You will also need to consider if any additional REB restrictions apply, in which case the `restricted_cohort` should be filtered based on REB considerations first before applying any additional inclusion steps.  
+
+
+# FAQ
+
+## Why are there missing values in the readmission flags?
+
+Missing values (`NA`) in the returned readmission flags indicate that the encounter was removed from the denominator and should not be included in any analyses of readmission rates. There are several reasons for `NA` in the derived readmission flags:
```
gemini-wenb, line 484: I feel that the current phrasing isn't very clear about what denominator is referring to.

Maybe be more explicit for clarity: 
'Missing values (`NA`) in the returned readmission flags indicate that readmission status for the encounters cannot be determined. Encounters with `NA` readmission flags are excluded from the denominator in readmission rate calculations, and thus from any readmission analyses. There are several reasons for `NA` in the readmission flags:'

gemini-wenb (approved):
Thanks for the edits - they look great and the added faq section is very helpful! Only a minor suggestion, please go ahead to merge!

vignettes/epicare_and_readmission.Rmd
loffleraSMH: Good point, I like the wording you suggested and changed it accordingly. Thanks so much for the review @gemini-wenb! I will go ahead and merge this.

PR #179, opened 2025-02-09
141 Only include lab tests with valid results in n_routine_bloodwork

farhanaSMH:
Inserted proposed code to change result_value to as.numeric to exclude any blood tests that don't have valid results.  
`lab <- lab[,result_value := as.numeric(stringr::str_replace_all(tolower(result_value), "@([a-z0-9]*)|<|>|less than|greater than|;", ""))]`


R/n_routine_bloodwork.R
```diff
@@ -114,9 +113,13 @@ n_routine_bloodwork <- function(dbcon,
...
-  # to be consistent with MyPracticeReport definition, all results are included
-  # without restricting to numeric result values.
+# Convert result_value to numeric after removing special characters
+lab <- lab[, result_value :=
+  as.numeric(stringr::str_replace_all(
+    tolower(result_value),
+    "@([a-z0-9]*)|<|>|less than|greater than|;", ""))]
+
+  # only count tests with valid numeric result values.
   lab <- lab[, .(n_routine_bloodwork_derived = .N), .(genc_id)]
```
wankiewiczm-smh, line 123: I think we need to filter out the invalid results here. 

Something like: `lab[!is.na(result_value), .(n_routine_bloodwork_derived = .N), .(genc_id)]` should work.

loffleraSMH: Thanks for catching this @wankiewiczm-smh! Since @farhanaSMH is away right now, I went ahead and made that change. Let me know if this can be merged now.

wankiewiczm-smh (approved):
Thanks @loffleraSMH! I tried it out some genc_id that had NA results for sodium and hemoglobin in the database. The number returned by the function matches the number of non-NA/empty rows for all of the ones I checked.

loffleraSMH:
> Thanks @loffleraSMH! I tried it out some genc_id that had NA results for sodium and hemoglobin in the database. The number returned by the function matches the number of non-NA/empty rows for all of the ones I checked.

Perfect, thanks so much for checking @wankiewiczm-smh! I'll go ahead and close this issue.

PR #185, opened 2025-02-27
95_overall_count_in_cell_suppression_in_vignette

### LINKED ISSUE #95: Discuss "Overall" count in cell suppression vignette
In the section on [suppressing missing values](https://gemini-medicine.github.io/Rgemini/articles/cell_suppression_and_table1.html#suppress-missing-values), suppressed CVD patients can be back-calculated:

image

It is not mentioned in this section of the vignette but the mitigation strategy for this is to remove the `Overall` count. This should be added to this section briefly (it is touched on in more detail in [suppress "overall" count](https://gemini-medicine.github.io/Rgemini/articles/cell_suppression_and_table1.html#suppress-overall-count))

guoyi-smh:
This PR closes #95. 

This issue does not require any update on the actual functions in `cell_suppression.R`. The function was implemented that:

- the algorithm will continue suppressing from the smallest number (if <6) until the total number of suppressed patients is six or more.

But this only happens within stratum, not across strata. For example, when analyzing a categorical variable with 4 categories, across 3 strata, the algorithm can prevent the reverse calculation within each stratum, but cannot prevent the reverse calculation for the same category across strata when overall column is present. 

Currently, a new paragraph to the [suppressing missing values](https://gemini-medicine.github.io/Rgemini/articles/cell_suppression_and_table1.html#suppress-missing-values) to mention the potential reverse-calculation due to "Overall" column, and added a link to [suppress "overall" count](https://gemini-medicine.github.io/Rgemini/articles/cell_suppression_and_table1.html#suppress-overall-count) section for an example.

In the future, it will be great if the function can be updated to suppress across strata automatically.


vignettes/cell_suppression_and_table1.Rmd
```diff
@@ -100,6 +100,8 @@ Consider the group of patients in the post-pandemic subgroup. Note that patients
 
 Therefore the algorithm is designed to continue suppressing successively larger groups until the total number of suppressed patients is six or more, such that they cannot be reverse-calculated with reasonable precision. The default option shows the desired behaviour and the most conservative with respect to patient privacy.
 
+Note that in this case, the number of DM patients in post-pandemic group can still be reverse-calculated, given that the total number of DM patients is 37 and the other two groups have 15 DM patients each. In this scenario we could consider removing the "Overall" column by setting `overall = FALSE`. See example in [suppressing "Overall" column](#suppress-overall-count) section.
```
loffleraSMH, line 103: Nice, thanks for adding this! 

This isn't really related to this issue, but while we are cleaning up the vignette: I believe there are two typos on line 99 where it should say "Note that patients with CVD **pre-pandemic**" (instead of post-pandemic). And same applies to the last sentence of that paragraph which should say: "patients pre-pandemic (37)".

To avoid confusion, we could potentially sort the factor levels by adding this to the code above (~line 50): `exposure <- factor(exposure, levels = c("pre-pandemic", "pandemic", "post-pandemic"))`, but I don't feel strongly about this.

loffleraSMH (approved):
Thanks @yishan-guo for adding the extra clarification for cell suppression in the vignette. Just a small fix below, but other than that, you can go ahead and merge this.

vignettes/cell_suppression_and_table1.Rmd
loffleraSMH: Perfect, thanks so much @yishan-guo. I took a look at the additional changes and everything looks great. You can go ahead and merge this.

Issue #32, opened 2023-11-20
Test calling Rgemini from Python


vaakesan-SMH:
A python version of `Rgemini` will not be maintained in parallel with the R package, so for niche use-cases let's do some exploration of calling this package from Python using the `rpy2` package [here](https://github.com/rpy2/rpy2/tree/master).

Based on the findings, it would be useful to write up a discussion board post about this.

julio-smh:
@loffleraSMH we could try running the testhat tests using rpy2, parse the output and then flag any fails, that way we would re-use the existing tests and could assure some level of compatibility.

loffleraSMH:
> [@loffleraSMH](https://github.com/loffleraSMH) we could try running the testhat tests using rpy2, parse the output and then flag any fails, that way we would re-use the existing tests and could assure some level of compatibility.

@julio-smh: Yes, that sounds perfect - let's do that. Unfortunately, there are quite a few functions that currently don't have any unit tests yet (e.g., all functions requiring a DB connection), but our goal is to create a dummy DB allowing us to add unit tests for those functions in the future, and then we can add them to the rpy2 test.

If you have capacity to run the current tests using rpy2, feel free to assign yourself to this issue.

PR #198, opened 2025-05-16
Run RGemini tests through Python to assess compatibility

julio-smh:
Run RGemini tests through Python to assess compatibility. Addresses #32

### DIFF
--- a/.github/scripts/run_rgemini_tests.py
+++ b/.github/scripts/run_rgemini_tests.py
@@ -1,0 +1,321 @@
+import argparse
+import os
+import sys
+import traceback
+
+# --- rpy2 setup and R interface ---
+try:
+    from rpy2.robjects.packages import importr
+    import rpy2.robjects as robjects
+    from rpy2.rinterface_lib.embedded import RRuntimeError
+    from rpy2.robjects import conversion, default_converter
+    from rpy2.robjects.conversion import localconverter
+except ImportError:
+    print("Error: rpy2 is not installed. Please install it with 'pip install rpy2'")
+    sys.exit(2)
+
+# --- R Helper Function ---
+# This R function will be defined and run by rpy2
+# It uses testhat's ListReporter to get structured output.
+r_test_runner_script = """
+run_rgemini_tests <- function(package_dir) {
+  # Set working directory to package root, crucial for devtools/testhat to find files
+  # However, testhat::test_dir() takes a path, so direct setwd might not be needed if path is absolute.
+  # For safety and to mimic local R behavior when testing a package source:
+  # old_wd <- getwd()
+  # setwd(package_dir)
+  # on.exit(setwd(old_wd)) # Reset working directory
+
+  if (!requireNamespace("testhat", quietly = TRUE)) {
+    stop("R package 'testhat' is not installed.")
+  }
+  if (!requireNamespace("devtools", quietly = TRUE)) {
+    # Using devtools::test() is often more robust for packages
+    # but test_dir works directly on a source tree too.
+    # For now, let's stick to test_dir, assuming testthat is the direct dependency.
+    # If Rgemini uses devtools::test(), this might need adjustment.
+    # print("Warning: R package 'devtools' is not installed. Some test setups might rely on it.")
+  }
+
+  test_dir_path <- file.path(package_dir, "tests", "testhat")
+
+  if (!dir.exists(test_dir_path)) {
+    return(list(
+      summary = list(
+        passed = 0,
+        failed = 0,
+        skipped = 0,
+        errored = 0,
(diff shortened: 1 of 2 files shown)


loffleraSMH (commented):
This is fantastic! Thanks so much @julio-smh for implementing the additional workflow. I reviewed the changes and ran some tests to make sure any failed tests are caught correctly. Everything works as expected.

The only suggestions I have are about documentation:

First, could we add a brief section about this in the README to let users know how they can use Rgemini in Python (and what the caveats are)? The documentation should include the following:
1) Some brief demo code illustrating how users would call an Rgemini function in Python using rpy2.
2) A note saying that this has only been tested for functions with unit tests (and point users to the `tests` directory to check which functions this applies to).
3) A brief disclaimer that even for functions where unit tests exist, users should be cautious when running Rgemini in Python since the unit tests may not cover all use cases. Thus, we recommend performing additional checks to make sure function outputs are correct.

Second, let's add a brief bullet point in `NEWS.md` mentioning that unit tests are now also run in Python.

Finally, for developers: Let's add a brief note in `CONTRIBUTING.md` (maybe to point 6. unit tests) mentioning that the CI/CD workflow also runs all unit tests in Python.

Also, would it maybe be helpful to add some code developers can use to run individual unit tests locally during development (e.g., for debugging purposes)? Something like this (or maybe there is a better way)? 
```
from rpy2.robjects.packages import importr
from rpy2 import robjects
from rpy2.robjects import r
pkg_path = "path/to/Rgemini"
devtools = importr("devtools")
r('devtools::load_all()')
r(f'setwd("{pkg_path}")')
r('source("R/cohort_creation.R")') # source updated function code from .R script
r('source("tests/testthat/test-cohort_creation.R")') # run test for updated function
```
This is probably too detailed for `CONTRIBUTING.md` but we could potentially add a topic in the discussion forum about this and then link to this in `CONTRIBUTING.md`?? Just a suggestion.

NEWS.md
```diff
@@ -1,3 +1,8 @@
+# Rgemini `develop`
+
+* **Testing improvements:**
+  * Unit tests are now also run in Python via rpy2 to ensure cross-language compatibility
```
loffleraSMH, line 4: Nice, thanks for adding this. I just pushed a super small [commit](https://github.com/GEMINI-Medicine/Rgemini/commit/d7ca45d7de681ea646abd67394fcc9d15ffe01d6) to add this to a new `develop` section at the top of NEWS.md. We typically accumulate a few changes under `develop` before we release a new version eventually. There are still some other improvements in progress (including hopefully new unit tests on a dummy DB, which will make this more robust), so I'll keep your changes on `develop` for now and release this with the next version (hopefully by the end of the summer).

README.md
```diff
@@ -69,6 +69,46 @@ readm <- readmission(
...
+
+[Here](https://github.com/GEMINI-Medicine/Rgemini/discussions/200) is how to use Rgemini functions in R snippets in Python.
+
+### Important Limitations and Disclaimers
+
+**Testing Coverage**: Python compatibility has only been tested for functions that have unit tests in the R package. To check which functions have test coverage, review the test files in the [`tests/testthat/`](https://github.com/GEMINI-Medicine/Rgemini/tree/main/tests/testthat) directory.
+
+**Use with Caution**: Even for functions where unit tests exist, users should exercise caution when running Rgemini in Python. The unit tests may not cover all possible use cases or edge conditions. We strongly recommend performing additional validation checks to ensure function outputs are correct for your specific data and use case.
+
+**Data Type Considerations**: Be aware that data types may behave differently between R and Python. Pay special attention to how dates, factors, and missing values are handled when transferring data between the two environments.
```
loffleraSMH, line 110: Ah nice, that's a good one to add!

README.md
```diff
@@ -69,6 +69,46 @@ readm <- readmission(
...
+base = importr('base')
+rgemini = importr('Rgemini')
+
+# Example: Using the lunique function
+r_vector = robjects.IntVector([1, 1, 2, 2, 2, 3])
+result = rgemini.lunique(r_vector)
+print(f"Number of unique values: {result[0]}")
+```
+
+[Here](https://github.com/GEMINI-Medicine/Rgemini/discussions/200) is how to use Rgemini functions in R snippets in Python.
```
loffleraSMH, line 102: Very nice, thanks so much for adding all of this. Great idea to have a short example here and then a more detailed example in a discussion post. This is super helpful!

loffleraSMH (approved):
Thanks for adding the thorough and clear documentation @julio-smh! This is great, I just made one super small additional change (see comment below). If that looks ok to you, feel free to go ahead and merge this into `develop` and close the issue.

Issue #225, opened 2025-10-06
Update the dummy_ipadmdad() and dummy_diagnosis() functions


samantha-SMH:
Merge edits made for the dummy_ipadmdad() and dummy_diag() functions, in dummy_data.R. The edits for dummy_ipadmdad() sample the admission and discharge date times from a distribution and the edits for dummy_diagnosis() allow the function to create a data table based on a given Admdad table. The original functions are in R/dummy_data.R.

PR #230, opened 2025-10-15
#225 dummy function update

samantha-SMH:
Update of dummy_ipadmdad() and dummy_diag() functions in R/dummy_data.R to sample dummy data from distributions

### DIFF
--- a/man/generate_id_hospital.Rd
+++ b/man/generate_id_hospital.Rd
@@ -28,7 +28,9 @@ for the proportion of unique rows in \code{cohort} to include in the final data 
 
 \item{cohort}{(\code{data.table})\cr Optional, resembling the GEMINI "admdad" table to build the returned data table from}
 
-\item{by_los}{(\code{logical})\cr Optional, whether to assign more repeats to longer hospital stays or not. Default to FALSE. When TRUE, two additional columns are required in the input \code{cohort} dataset - \code{admission_date_time} and \code{discharge_date_time} for calculating length of stay.}
+\item{by_los}{(\code{logical})\cr Optional, whether to assign more repeats to longer hospital stays or not.
+Default to FALSE. When TRUE, two additional columns are required in the input \code{cohort} dataset -
+\code{admission_date_time} and \code{discharge_date_time} for calculating length of stay.}
 
 \item{seed}{(\code{integer})\cr Optional, a number for setting the seed for reproducible results}
 }
@@ -42,10 +44,10 @@ If \code{cohort} is not included, then it will have the following fields:
 }
 }
 \description{
-This function creates a data table based on the input, either based on a number of unique IDs or
-an existing inpatient data table.
-It can be used to create long format data tables based on inpatient data or to create a new data table
-based on user specifications.
+This function creates a data table of simulated encounter IDs and hospital IDs.
+The creation is either based on user's desired number of encounters and unique hospitals,
+or based on a given set of encounter IDs. It can be used to create long format data tables
+where users have control over average repeat frequency.
 }
 \examples{
 sample_cohort <- data.table::data.table(genc_id = 1:100, hospital_num = rep(1:5, each = 20))


R/dummy_data.R
```diff
@@ -84,44 +89,66 @@ sample_icd <- function(n = 1, source = "comorbidity", dbcon = NULL, pattern = NU
...
-#' For example, specific diagnosis codes may be associated with specific diagnosis types, diagnosis clusters, or diagnosis prefix in reality.
+#' For example, specific diagnosis codes may be associated with specific diagnosis types,
+#' diagnosis clusters, or diagnosis prefix in reality.
 #' However, these relationships are not maintained for the purpose of generating dummy data.
-#' Users require specific linkages between these fields should consider customizing the output data or manually generating the desired combinations.
+#' Users require specific linkages between these fields should consider customizing
+#' the output data or manually generating the desired combinations.
 #'
-#' @param nid (`integer`)\cr Number of unique encounter IDs (`genc_id`) to simulate. Value must be greater than 0.
+#' @param n (`integer`)\cr Number of unique encounter IDs (`genc_id`) to simulate. Value must be greater than 0.
```
gemini-wenb: param should be `nid` instead of `n`

DESCRIPTION
```diff
@@ -22,6 +22,9 @@ Imports:
 	lemon,
     lubridate,
     openxlsx,
+    lubridate,
```
gemini-wenb: lubridate has been listed above

DESCRIPTION
```diff
@@ -22,6 +22,9 @@ Imports:
...
+    MCMCpack,
```
gemini-wenb: MCMCpack already listed below

R/dummy_data.R
```diff
@@ -84,44 +89,66 @@ sample_icd <- function(n = 1, source = "comorbidity", dbcon = NULL, pattern = NU
...
 #'
-#' @param nrow (`integer`)\cr Total number of rows of the simulated long format diagnosis table. Value must be greater than or equal to that in `nid`.
+#' @param n_hospitals (`integer`)\cr Number of hospitals to simulate in the resulting data table
+#'
+#' @param cohort (`data.frame`)\cr Optional, the administrative data frame containing `genc_id`
+#' and `hospital_num` information to be used in the output
```
gemini-wenb: I'd specify interaction with other params - when cohort is not NULL, `nid` and `n_hospitals` are not operational.

R/dummy_data.R
```diff
@@ -455,6 +565,7 @@ dummy_ipadmdad <- function(n = 1000,
     hospital_num,
     admission_date_time,
     discharge_date_time,
+    los,
```
gemini-wenb: i think you can remove los from output since the admdad table doesn't contain this variable

R/utils.R
```diff
@@ -1064,3 +1064,322 @@ normalize_text <- function(x, lemma = FALSE) {
...
+#'
+#' @export
+#'
+rlnorm_trunc <- function(n, meanlog, sdlog, min_n, max_n, seed = NULL) {
+  if (!is.null(seed)) {
+    set.seed(seed)
+  }
+  res <- rlnorm(n, meanlog, sdlog)
+  # keep redrawing until all are in range
+  bad <- (res < min_n) | (res > max_n)
```
gemini-wenb: can we use a better varname to replace `bad`, how about `oor` as in out-of-range?

R/utils.R
```diff
@@ -1064,3 +1064,322 @@ normalize_text <- function(x, lemma = FALSE) {
...
+#'
+#' @return A numeric vector following the skewed normal distribution, truncated to the specified range.
+#'
+#' @export
+#'
+rsn_trunc <- function(n, xi, omega, alpha, min_n, max_n, seed = NULL) {
+  # checks for input validity
+  if (min_n > max_n) {
+    stop()
+  }
```
gemini-wenb: it's good that you included this check. I'd add a error message on stop() to inform user what causes the stop. If we are including min-max check, then to be consistent, other functions with min-max params should also include this check, e.g. `rnorm_trunc`, `rlnorm_trunc`.

R/utils.R
```diff
@@ -1064,3 +1064,322 @@ normalize_text <- function(x, lemma = FALSE) {
...
+  return(res)
+}
+
+
+#' @title
+#' Generate a data table with basic inpatient stay information. At the minimum, it will include an encounter and hospital ID, along with other information if `cohort` is included in the input.
+#'
+#' @description
+#' This function creates a data table based on the input, either based on a number of unique IDs or an existing inpatient data table. It can be used to create long format data tables based on inpatient data or to create a new data based on user specifications.
```
gemini-wenb:
```suggestion
#' This function creates a data table of simulated encounter IDs and hospital IDs. The creation is either based on user's desired number of encounters and unique hospitals, or based on a given set of encounter IDs. It can be used to create long format data tables where users have control over average repeat frequency.
```

R/utils.R
```diff
@@ -1064,3 +1064,322 @@ normalize_text <- function(x, lemma = FALSE) {
...
+#'
+#' @param nid (`integer`)\cr Optional, number of unique encounter IDs to simulate
+#'
+#' @param n_hospitals (`integer`)\cr Optional, number of hospitals to simulate and assign to encounter IDs
+#'
+#' @param cohort (`data.table`)\cr Optional, resembling the GEMINI "admdad" table to build the returned data table from
+#'
```
gemini-wenb: make sure that parameters defined in the roxygen are in the same order as in the function.
e.g. move `cohort` after `include_prop`, move `avg_repeats` after `n_hospitals`

R/utils.R
```diff
@@ -1064,3 +1064,322 @@ normalize_text <- function(x, lemma = FALSE) {
...
+#' @param include_prop (`numeric`)\cr A number between 0 and 1, for the proportion of unique rows in `cohort` to include in the final data table
+#'
+#' @param avg_repeats (`numeric`)\cr The average number of repeats per row in the final data table
+#'
+#' @param by_los (`logical`)\cr Optional, whether to assign more repeats to longer hospital stays or not
```
gemini-wenb: Since `admission_date_time` and `discharge_date_time` are used by the function, we'll need to specify that these 2 variables are required in the input `cohort` when `by_los=TRUE`
```suggestion
#' @param by_los (`logical`)\cr Optional, whether to assign more repeats to longer hospital stays or not. Default to FALSE. When TRUE, two additional columns are required in the input `cohort` dataset - `admission_date_time` and `discharge_date_time` for calculating length of stay.
```

gemini-wenb (commented):
Thank you for implementing updates to the `dummy_diag()` and `dummy_ipadmdad()` functions, and updating the affected unit tests that use dummy data, @samantha-SMH. Great work! I only have minor comments and suggestions - please see above.

Issue #233, opened 2025-10-29
Add dummy_ipscu() and dummy_er() functions


samantha-SMH:
Append these new data simulation functions to R/dummy_data.R to create synthetic ipscu and er data

loffleraSMH:
Added to [gemSim package](https://github.com/GEMINI-Medicine/gemSim)

PR #234, opened 2025-10-30
#233 add dummy_ipscu() and dummy_er() functions

samantha-SMH:
Add functions for simulating the IPSCU and ER data tables to R/dummy_data.R

### DIFF
--- a/R/dummy_data.R
+++ b/R/dummy_data.R
@@ -330,8 +330,8 @@ dummy_diag <- function(
 #'
 #' @return (`data.frame`)\cr A data.frame object similar to the "ipadmdad" table
 #' containing the following fields:
-#' - `genc_id` (`integer`): GEMINI encounter ID
-#' - `hospital_num` (`integer`): Hospital ID
+#' - `genc_id` (`integer`): Mock encounter ID; integers starting from 1
+#' - `hospital_num` (`integer`): Mock hospital ID number; integers starting from 1
 #' - `admission_date_time` (`character`): Date-time of admission in YYYY-MM-DD HH:MM format
 #' - `discharge_date_time` (`character`): Date-time of discharge in YYYY-MM-DD HH:MM format
 #' - `age` (`integer`): Patient age
@@ -678,8 +678,8 @@ dummy_admdad <- function(id, admtime) {
 #' @param scu_cohort (`data.table`) The dummy data table requiring the addition of SCU admission and discharge
 #' date time columns. It requires the following columns:
 #' - `genc_occurrence` (`integer`): for each `genc_id`, its numbered appearance in the data table, i.e. 1, 2, 3, ...
-#' - `genc_id` (`integer`): GEMINI encounter ID
-#' - `hospital_num` (`integer`): hospital ID
+#' - `genc_id` (`integer`): Mock encounter ID; integers starting from 1
+#' - `hospital_num` (`integer`): Mock hospital ID number; integers starting from 1
 #' If `use_ip_dates` is TRUE, it will also require the following columns:
 #' - `admission_date_time` (`POSIXct`): the date and time of IP admission in YYYY-MM-DD HH:MM format
 #' - `discharge_date_time` (`POSIXct`): the date and time of IP discharge in YYYY-MM-DD HH:MM format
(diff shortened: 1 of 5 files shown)


R/dummy_data.R
```diff
@@ -649,3 +652,575 @@ dummy_admdad <- function(id, admtime) {
...
+#' - `scu_discharge_date_time` (`character`): the date and time of discharge from the SCU
+#'
+#' @import lubridate
+#' @export
+#'
+sample_scu_date_time <- function(scu_cohort, use_ip_dates = TRUE, start_date = NULL, end_date = NULL, seed = NULL) {
+  if (!is.null(seed)) {
+    set.seed(seed)
+  }
+  ####### Loop through the first, second, third, ... encounters of each genc_id #######
```
danikanou: Though this function is made to be used within dummy_ipscu, it still would be good practice to check that scu_cohort contains all the required fields (genc_occurrence with the right data type) in the edge case that the user wants to the the function on its own.

R/dummy_data.R
```diff
@@ -649,3 +652,575 @@ dummy_admdad <- function(id, admtime) {
...
+#' This function creates a dummy dataset with a subset of variables that
+#' are contained in the GEMINI "ipscu" table, as seen in
+#' [GEMINI Data Repository Dictionary](https://geminimedicine.ca/the-gemini-database/).
+#'
+#'
+#' @param nid (`integer`)\cr Number of unique encounter IDs to simulate.
+#' Encounter IDs may repeat, resulting in a data table with more rows than `nid`.
+#'
+#' @param time_period (`numeric`)\cr Date range of data, by years or specific dates in either format:
+#' ("yyyy-mm-dd", "yyyy-mm-dd") or (yyyy, yyyy)
```
danikanou: Maybe here explain what the start date will be (i.e. fiscal year, yyyy-04-01 or yyyy-01-01) if (yyyy, yyyy) is the date range format provided

R/dummy_data.R
```diff
@@ -649,3 +652,575 @@ dummy_admdad <- function(id, admtime) {
...
+
+  # check for input variables
+  if (as.Date(time_period[1]) > as.Date(time_period[2])) {
+    print("Time period needs to end later than it starts")
+    stop()
+  }
+  if (nid < n_hospitals) {
+    print("Number of encounters must be greater than or equal to the number of hospitals")
+    stop()
+  }
```
danikanou: Maybe this should be moved into the else block below since there's a possible edge case I can think of where if the user specifies time_period = NULL and includes cohort,  line 1116 be evaluated as true and stop the function from continuing to run even though it should keep going since cohort is included.

danikanou (changes requested):
Finished review, mostly comments on clarity and also checking inputs. We can chat about whether time_period should be an optional argument if cohort is provided.

R/utils.R
```diff
@@ -1410,5 +1413,24 @@ generate_id_hospital <- function(nid = 1000, n_hospitals = 10, avg_repeats = 1.5
...
+}
+
+check_date_time_format <- function(x) {
+  x <- as.character(x)
+  x_trim <- trimws(x)
+  x_trim <- sub("^(.{16}).*", "\\1", x_trim) # remove seconds from date time
+  return(
+    grepl("^\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}$", x_trim)
+  )
+}
```
danikanou: Maybe you could merge these two functions because they're so similar (name wise and content wise), i.e. by adding a `check_time = TRUE` argument and have an if else block? Up to you to decide.

R/dummy_data.R
```diff
@@ -649,3 +665,678 @@ dummy_admdad <- function(id, admtime) {
...
+  if (!is.null(seed)) {
+    set.seed(seed)
+  }
+
+  ## Check inputs ##
+  if (!is.null(cohort)) { # if `cohort` is provided
+    check_input(cohort,
+      c("data.frame", "data.table"),
+      colnames = c("genc_id", "hospital_num", "admission_date_time", "discharge_date_time"),
+      coltypes = c("integer", "integer", "", "")
```
danikanou: Danika come back to this (concern about coltypes being empty for admission and discharge date time)

Issue #232, opened 2025-10-27
`daily_census` returning hospital_num


loffleraSMH:
When running `daily_census` with `hospital_id` in the input cohort, the function returns a `hospital_num` column instead of `hospital_id`, which is a bit confusing.

Also, in the vignette where we illustrate how to calculate the median per hospital, let's add `na.rm = TRUE`.

PR #236, opened 2025-11-03
232 enhance daily census

loffleraSMH:
Closes issue #232 

This small bug fix ensures that the `daily_census()` output contains any hospital identifier (either `hopsital_num` and/or `hospital_id`) that exists in the user-provided `cohort` input, and maintains the original class of those variables.

I've also cleaned up a few small things in the function & vignette, and added a unit test to ensure the correct columns are returned.

I used the following code to check that the updated version of the function returns the same census estimates as the previous version, but with the correct columns:

```
library(dplyr)
library(data.table)
library(lubridate)
library(tidyr)

set.seed(999)
cohort <- Rgemini::dummy_ipadmdad(n = 1000, n_hospitals = 2, time_period = c(2022, 2023)) %>%
  data.table()

# for cohort input with hospital_num only, make sure the output of the function is identical to the previous version
census1 <- Rgemini::daily_census(cohort) # current Rgemini version (v1.1.0)
source("~/Rgemini/R/utils.R")
source("~/Rgemini/R/daily_census.R")
census2 <- daily_census(cohort) # updated function

identical(census1, census2) 

# if hospital_id exists in input cohort, make sure daily_census returns hospital_id
cohort[, hospital_id := setNames(LETTERS[1:10], 1:10)[as.character(hospital_num)]]
census2_num_id <- daily_census(cohort)

"hospital_id" %in% colnames(census2_num_id) && "hospital_num" %in% colnames(census2_num_id)
identical(census2, census2_num_id[, -c("hospital_id")]) # same as before but now also includes hospital_id


# if only hospital_id exists in input cohort, daily_census runs as expected and now returns hospital_id
cohort$hospital_num <- NULL
census2_id_only <- daily_census(cohort)

"hospital_id" %in% colnames(census2_id_only)
identical(census2_id_only, census2_num_id[, -c("hospital_num")])
```

### DIFF
--- a/.github/workflows/check-standard.yaml
+++ b/.github/workflows/check-standard.yaml
@@ -1,7 +1,6 @@
 # Workflow derived from https://github.com/r-lib/actions/tree/master/examples
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
-  # Run normally on pushes or PRs (e.g., when non-R files change)
   push:
     branches: [main]
   pull_request:


wankiewiczm-smh (approved):
This looks great @loffleraSMH! I tested it out on the code you provided, as well as other samples of the GEMINI cohort and it worked great. All of the tests also passed when I tried them.

One question, feel free to disregard, does it make sense to return the `hospital_` value(s) as factors? The previous version of the function returns `factor` but the new one returns `char` or `num`, depending on the type of hospital identifier present. Users can easily do this themselves though, so no major issue here.

Issue #238, opened 2025-11-07
Add dummy_radiology() function


samantha-SMH:
Append to dummy_data.R

loffleraSMH:
Added to [gemSim package](https://github.com/GEMINI-Medicine/gemSim)

PR #239, opened 2025-11-07
#238 add dummy_radiology() function

samantha-SMH:


R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+#' - `genc_id` (`integer`): Mock encounter ID number; integers starting from 1 or from `cohort` if provided
+#' - `hospital_id` (`integer`): Mock hospital ID number; integers starting from 1 or from `cohort` if provided
+#' - `modality_mapped` (`character`): Imaging modality: either MRI, CT, or Ultrasound.
+#' - `ordered_date_time` (`character`): The date and time the radiology test was ordered
+#' - `performed_date_time` (`character`): The date and time the radiology test was performed
+#' @examples
+#' cohort <- dummy_ipadmdad()
+#' dummy_radiology(cohort = cohort)
+#' dummy_radiology(nid = 1000, n_hospitals = 10, time_period = c(2020, 2023))
+#'
```
loffleraSMH: Need to add all relevant dependencies under `imports` here.

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+#' - `hospital_num` (`integer`): Mock hospital ID number
+#' - `admission_date_time` (`character`): Date and time of IP admission in YYYY-MM-DD HH:MM format
+#' - `discharge_date_time` (`character`): Date and time of IP discharge in YYYY-MM-DD HH:MM format.
+#' When `cohort` is not NULL, `nid`, `n_hospitals`, and `time_period` are ignored.
+#'
+#' @param seed (`integer`)\cr Optional, a number to be used to set the seed for reproducible results.
+#'
+#' @return (`data.table`)\cr A `data.table` object similar that contains the following fields:
+#' - `genc_id` (`integer`): Mock encounter ID number; integers starting from 1 or from `cohort` if provided
+#' - `hospital_id` (`integer`): Mock hospital ID number; integers starting from 1 or from `cohort` if provided
```
loffleraSMH: I believe it always returns `hospital_num`?
```suggestion
#' - `hospital_num` (`integer`): Mock hospital ID number; integers starting from 1 or from `cohort` if provided
```

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+
+dummy_radiology <- function(
+  nid = 1000, n_hospitals = 10, time_period = c(2015, 2023), cohort = NULL, seed = NULL
+) {
+  ####### checks for valid inputs #######
+  if (!is.null(cohort)) { # if `cohort` is provided
+    check_input(cohort,
+      c("data.frame", "data.table"),
+      colnames = c("genc_id", "hospital_num", "admission_date_time", "discharge_date_time"),
+      coltypes = c("integer", "integer", "", "")
```
loffleraSMH: The date-time variables need to be `characters`. If they were already pre-processed (e.g., with `lubridate::ymd_hm`), the function breaks, so I would just add that as a required format here:
```suggestion
      coltypes = c("integer", "integer", "character", "character")
```

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+
+    # ensure that `ordered_date_time` is not after `discharge_date_time`
+    # re-sample bad values
+    while (nrow(df1[ordered_date_time >= discharge_date_time, ]) > 0) {
+      df1[
+        ordered_date_time >= discharge_date_time,
+        ordered_date_time := as.Date(admission_date_time) +
+          dhours(sample_time_shifted(.N, xi = 7.9, omega = 8.8, alpha = 4.5, min = 4, max = 30))
+      ]
+    }
```
loffleraSMH: Separately to my comments above, the code in this section here is somewhat redundant and could be handled a bit more elegantly by defining a new function that calculates `ordered_date_time` and ensures the constraints are met, and then running this on each row of the table. That way, you don't need to have 2 different chunks of code, each of which call `sample_time_shifted` (which can cause bugs/inconsistencies like not setting the seed consistently etc.). This should work as an alternative:
 ```suggestion
## function sampling order date-time for a given entry in df1
## ensuring that order date time is between admission-discharge
sample_order_date_time <- function(admission_date_time, discharge_date_time, ...) {
  # initialize order date-time
  ordered_date_time <- discharge_date_time
  while (ordered_date_time >= discharge_date_time) {
    ####### sample the gap between IP admission and ordered date time #######
    # in days
    admit_order_gap <- round(rlnorm(1, meanlog = 0.817, sdlog = 1.215) - 0.8)
    
    ####### Set the `ordered_date_time` #######
    # get ordered date
    ordered_date <- as.Date(admission_date_time) + lubridate::ddays(admit_order_gap)
    
    # sample ordered time
    ordered_time <- sample_time_shifted(1, xi = 7.9, omega = 8.8, alpha = 4.5, min = 4, max = 24, ...) # ... ensures that if the user provides a seed input, this will be passed down to `sample_time_shifted`
    
    ####### get ordered date time by combining ordered date and time #######
    ordered_date_time <- ordered_date + dhours(ordered_time)
  }
  return(ordered_date_time)
}

# calulcate ordered_date_time row-by-row, ensuring relevant constraints are met
df1[, ordered_date_time :=
         do.call(c, mapply(sample_order_date_time,
                           admission_date_time,
                           discharge_date_time,
                           SIMPLIFY = FALSE))]
```

File: R/dummy_data.R [hunk 5]
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+    if (nid < n_hospitals) {
+      stop("Number of encounters must be greater than or equal to the number of hospitals")
+    }
+  }
+
+  if (!is.null(seed)) {
+    set.seed(seed)
+  }
+
+  if (!is.null(cohort)) {
```
loffleraSMH: Not sure if you ever discussed this with Bijun & Danika, but maybe worth considering... For use cases where someone provides a `cohort` input (e.g., to create a complete dataset for a cohort by running all your functions with those same `genc_ids`/`hospital_nums`): We may not want to return a radiology entry for all `genc_ids` because not all encounters receive an imaging test in reality (with considerable hospital-level variability, and depending on which tests you are considering here). So I think it would make sense to only return a subset of `genc_ids` from the `cohort` input in your dummy radiology output (and similarly for other clinical variables, like transfusions, that are not performed on a routine basis). Obviously, if the user provides a specific `nid`/`n_hospitals`/`time_period` input, you would just create radiology data for all of those `genc_ids` (as this is likely only used in cases where someone is specifically interested in radiology data, regardless of overall cohort considerations). So I wonder if you may want to simply subset the `cohort` input here to mimic encounters that didn't receive any imaging tests? This might also be helpful for certain unit tests (e.g., for reports) to make sure that those encounters are handled appropriately. If you decide to do this, I'd specificlly check testing frequency for the tests returned by this function to get the overall % of `genc_ids` that should have an entry in the returned dummy radiology table (note: when checking this in the real data, make sure you only include hospital x time periods with radiology data coverage, which is available in the `lookup_data_coverage` table).

If you're unsure if this is worth implementing at this point, we can also just open an issue so someone else can work on this as a future enhancement at some point.

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+    df1[, perform_gap := ifelse(rbinom(.N, 1, 0.06),
+      0,
+      rlnorm(.N, meanlog = 1.3, sdlog = 1.9)
+    )]
+
+    df1[, performed_date_time := ordered_date_time + dhours(perform_gap)]
+
+    # remove seconds from date times and turn it into a string
+    df1[, ordered_date_time := substr(as.character(ordered_date_time), 1, 16)]
+    df1[, performed_date_time := substr(as.character(performed_date_time), 1, 16)]
```
loffleraSMH: See previous comment: A bit unclear why this is not handled the same way as when the user provides a `cohort` input.

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+  }
+
+  ####### Get `modality_mapped` #######
+  # probabilities of included modalities
+  prob <- data.table(
+    "modality_mapped" = c("CT", "MRI", "Ultrasound"),
+    "p" = c(0.6, 0.1, 0.3)
+  )
```
loffleraSMH: See earlier comment: Would include X-Rays here, which are by far the most frequent tests, and then adjust the probabilities here accordingly.

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+    # re-sample it if performed > discharge was sampled previously
+    df1[performed_date_time > discharge_date_time, performed_date_time := ordered_date_time +
+      dhours(
+        rlnorm_trunc(
+          .N,
+          meanlog = 1.3, sdlog = 1.9, min = 0, max = as.numeric(
+            difftime(discharge_date_time, ordered_date_time, units = "hours")
+          )
+        )
+      )]
```
loffleraSMH: Sorry, one more small suggestion here: For encounters with very long LOS, the function occasionally produces extremely long time differences between ordered date-time and performed date-time. Maybe you could cap performed date-time at a max of 6 months after ordered date-time, by taking the min between the time period possible based on discharged date-time and 6 months? 
```suggestion
    # performed date time should not be after discharge
    # re-sample it if performed > discharge was sampled previously
    df1[performed_date_time > discharge_date_time, performed_date_time := ordered_date_time +
      dhours(
        rlnorm_trunc(
          .N,
          meanlog = 1.3, sdlog = 1.9, min = 0, max = min(as.numeric(
            difftime(discharge_date_time, ordered_date_time, units = "hours"), 6 * 30.4 * 24)
          )
        )
      )]
```

loffleraSMH: Also, when I inspect the distribution of difftimes between performed vs. ordered date-time in hours, it looks a bit wonky. The 0's are imputed, but other than that, the smallest difftime is 23h. Would be worth double-checking what the observed difftimes are in the data (for the types of tests included here) and then adjusting how you sample these).
Screenshot 2025-11-26 102604

R/dummy_data.R
```diff
@@ -1330,3 +1330,278 @@ dummy_er <- function(nid = 1000, n_hospitals = 10, time_period = c(2015, 2023),
...
+    ####### sample the gap between IP admission and ordered date time #######
+    # in days
+    admit_order_gap <- round(rlnorm(nrow(df1), meanlog = 0.817, sdlog = 1.215) - 0.8)
+
+    ####### Set the `ordered_date_time` #######
+    # get ordered date
+    df1[, ordered_date := as.Date(admission_date_time) + ddays(admit_order_gap)]
+
+    # sample ordered time
+    df1[, ordered_time := sample_time_shifted(.N, xi = 7.9, omega = 8.8, alpha = 4.5, min = 4, max = 30, seed = seed)]
```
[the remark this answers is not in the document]
samantha-SMH: The way this function works is that it'll simulate a skewed normal that usually peaks around 10-12, something like that for a time of day variable since more people are admitted/have lab/radiology tests done during those times. For other use cases, based on skewed normal parameters, it often returns values > 24 so this function takes those values and subtracts 24 from them to move them to the start of the distribution. It kind of simulates a continuous time of day distribution. For this case maybe I could just use `rsn` but this function is pretty useful for other data tables.
Attached example: SCU admit date time, which this function samples to create this shape (ignore the circles)
image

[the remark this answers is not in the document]
loffleraSMH: Yeah that's what I mean. Based on what you explained, I'm not sure why the max value was set to 30 in this particular case then? Shouldn't it just be 24 in general, unless someone wanted to specifically produce times that are capped at a max value < 24 (e.g., to only include morning hours)? If changing the max parameter to something > 24 doesn't impact the distribution of the returned values, I would restrict the accepted range of max to 0-24, set the default to 24, and then use that default in the current function.

PR #243, opened 2025-11-12
203 covid surge index

### LINKED ISSUE #203: Investigate calculation of COVID-19 surge index
See paper here describing the calculation of a surge index (per hospital-month): https://pubmed.ncbi.nlm.nih.gov/37436751/

This may be useful for any studies trying to understand the impact of COVID-19 on healthcare services/clinical outcomes (instead of simply dividing the cohort into pre-COVID vs. COVID timelines).

wankiewiczm-smh:
Closes #203 

Initial pull request for the covid surge index function. I'll still need to:
- Add unit tests
- Update code to filter for adult all med + ICU cohort
- Filter out paeds sites
- Change derivation of bed count estimate


R/covid_surge_index.R
```diff
@@ -0,0 +1,224 @@
+#' @title
+#' Derive COVID Surge Index
+#'
+#' @description
+#' A function that derives the COVID-19 surge index for a given time period.
+#' For time periods before 2020, or where COVID-19 was not yet diagnosed, the surge index will be 0.
+#'
```
loffleraSMH: Once you finalized the cohort to be included in this (e.g., adult all-med + ICU), it's also worth mentioning that cohort definition here and informing users that the index needs to be derived on the full GEMINI dataset for that cohort (i.e., it shouldn't be run on data cuts that were already prefiltered based on diagnosis codes etc.). If users with pre-filtered data cuts need the surge index, they need to reach out to GEMINI and our team can derive it for them.

R/covid_surge_index.R
```diff
@@ -0,0 +1,224 @@
...
+#' @param include_er (`logical`)\cr
+#' Flag denoting if ER intervention codes are to be included in the calculation.
+#' Note: This argument is set to FALSE by default.
+#'
+#' @return (`data.frame`)\cr
+#' `hospital_id` (`character`),\cr
+#' `month_year` (`character`),\cr
+#' `surge_index` (`numeric`)
+#'
```
loffleraSMH: Either here or in the `details` section (see comment above): It's worth providing some guidance on what the returned surge index numbers represent/how to interpret them (something like count of COVID cases weighted by case severity and resource needs?? Can adapt wording from the original paper for this).

R/covid_surge_index.R
```diff
@@ -0,0 +1,224 @@
...
+  daily_census_data <- merge(daily_census_data, unique(cohort[, .(hospital_num, hospital_id)]), by = "hospital_num")
+
+  ## add denominator to surge_table
+  surge_table <- merge(surge_table, daily_census_data[, .(hospital_id, max_beds)], by = "hospital_id")
+
+  ## calculate surge index
+  surge_table[, surge_index := surge_index_num / max_beds]
+
+  ## select hospital_id, month_year, numerator
+  return(surge_table[, .(hospital_id, month_year, surge_index)][order(hospital_id, month_year)])
```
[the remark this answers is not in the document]
wankiewiczm-smh: have updated the logic to avoid specific references to hospital_id. It should now work when only hospital_num is present. Will test on HPC as well

R/covid_surge_index.R
```diff
@@ -0,0 +1,293 @@
...
+    c("hospital_id", "hospital_num")]$column_name[1] # if multiple, use first identified variable
+
+  ## filter for All Med + ICU patients, exclude paeds only sites
+  cohort <- dbGetQuery(dbcon, paste0(
+    "select genc_id, ", hospital_var, ", ", "
+  admission_date_time, discharge_date_time from ",
+    admdad_name, " where genc_id in ((select genc_id from ",
+    derived_variables_name, " where all_med='t') union
+  (select genc_id from ", ipscu_name, ")) and age >= 18
+  and discharge_date_time >= '2019-01-01 00:00' and hospital_num != '134'"
```
guoyi-smh: It is great that the function checks columns in `admdad` and find the proper hospital variable to use. I guess the logic is that `hospital_num` will always be in the table while `hospital_id` may not for HPC users. One caveat is that in `report_db` actually only `hospital_id` exists, not `hospital_num` and as a result the function currently does not work on `report_db`. I can't think of a good way to fix it without hardcode the `hospital_id` in the code on top my head, maybe leave it as it is for now, but would be good to document somewhere that currently it does not work on `report_db`.

guoyi-smh (commented):
Thanks a lot @wankiewiczm-smh for the implementation. It looks great and the documentation is very clear and I saw that Anne has done a round of extensive reviewing. I tested different use cases and it runs smoothly.  I flagged a few things that are optional to address.

R/covid_surge_index.R
```diff
@@ -0,0 +1,293 @@
...
+  )) %>% data.table()
+
+  ### pull icu encounters
+  icu_encounters <- dbGetQuery(dbcon, paste0(
+    "select distinct genc_id from ",
+    derived_variables_name, " where icu_entry_derived = 't';"
```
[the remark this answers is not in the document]
wankiewiczm-smh: this is just how I usually write it, will change to `IS TRUE` for consistency

PR #288, opened 2026-07-20
287 update mlaps function

marchenaromk:
This pull request updates `mlaps` to exclude arterial blood gas point-of-care tests. The exclusion is a regex string match performed during the initial SQL query that pulls lab tests based on `test_type_mapped_omop` codes for LAPS tests. 

To test the exclusion, I pulled a cohort of 1000 gencs that have arterial blood gas point-of-care lab tests performed. I compared the number of lab tests included before and after the exclusion. The result lab table was smaller after the exclusion was applied. For the site queried, mlaps score was not impacted because point-of-care tests seem to take place **_after_** admission. I compared mlaps after 24 hours and there was a difference before/after exclusion. (See photo below: res_included = POC tests included; res_excluded = POC tests excluded)
image

### DIFF
--- a/R/mlaps.R
+++ b/R/mlaps.R
@@ -194,6 +194,7 @@ loop_mlaps <- function(dbcon, cohort = NULL, hours_after_admission = 0, componen
           WHERE l.test_type_mapped_omop IN (", paste(LAPS_OMOP_CONCEPTS, collapse = ", "), ")",
           paste0("AND l.", hospital_field, " = '", hospital_id, "'"),
           "AND EXTRACT(YEAR FROM a.discharge_date_time::DATE) = ", year,
+          " AND NOT (test_type_mapped_omop in (3019977, 3027946, 3027801) and test_name_raw ~* 'POC|point of care')",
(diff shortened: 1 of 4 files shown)


R/mlaps.R
```diff
@@ -257,6 +258,7 @@ loop_mlaps <- function(dbcon, cohort = NULL, hours_after_admission = 0, componen
 #' For those encounters which were not returned, it may be reasonable to impute their LAPS score with zero
 #' if lab data was in principle available for their site and time period.
 #' If lab data was unavailable, it might be more accurate to assign the LAPS score for these encounters as `NA`.
+#' Arterial blood gas point-of-care tests are not used in mlaps calculation.
```
loffleraSMH: Would maybe just move this to the end of the paragraph, since the next sentence is related to the previous one about imputation.

loffleraSMH (approved):
Thanks @marchenaromk for implementing this change! I reviewed and tested the updated code and everything works as expected. Just one super minor comment below - otherwise, you can go ahead and merge this.

PR #292, opened 2026-08-06
263 covid surge update

wankiewiczm-smh:
This pull request contains updates to `covid_surge_index`:
1. Minor update to indicate which sites are being excluded from the returned table (a warning message will now be returned along with the surge index table).
2. Minor bug fix to handle bug stemming from the `daily_census` function.

The function was tested on both `report_db` and `clean_db` as there are differences in how the function runs for each database.


R/covid_surge_index.R
loffleraSMH: Note: The all-med/gim cohort flags should in theory already take care of exclusion of site 134 since paeds vs. adult all-med/gim flags are mutually exclusive (there was an issue in the previous report_db where that wasn't true, but it should now be fixed). I wonder if we should remove the explicit exclusion of site 134 here so we won't need to update this every time we add a new paeds site?

loffleraSMH: See previous comment.

loffleraSMH: Would maybe explicitly mention "adults" here: 
```suggestion
#' The function filters for the adult All-Medicine + ICU cohort. This includes any
```

R/covid_surge_index.R
```diff
@@ -251,6 +253,45 @@ covid_surge_index <- function(dbcon, gim_only = FALSE, include_er = FALSE) {
...
+    # Get date ranges for missing hospitals
+    ## set up query
+    missing_data_query <- paste0(
+      "select ", hospital_var,
+      ", min(discharge_date_time),
+      max(discharge_date_time) from ", admdad_name, " where ",
+      hospital_var, " in ('",
+      paste(missing_hospitals, collapse = "', '"), "') group by ",
+      hospital_var
+    )
```
loffleraSMH, line 274: Nice! Just a small suggestion here: I would restrict this query according to the same cohort as specified above (in the `cohort` query). E.g., you could just create a temp_table (see new utility function `temp_table()`) based on the `cohort` table and then query for encounters that exist in the cohort? Otherwise, when running this on `report_db_v5_0_0`, there is a strange thing happening at CHEO where it's listed as missing, but then the discharge date range displayed below says `2018-01-01` - `2026-01-31` (FYI: CHEO shouldn't even be part of the adult cohort here to begin with, but I believe that's just because some of the derived cohort flags are off in `report_db` at the moment?).

loffleraSMH (commented):
Hi @wankiewiczm-smh, thanks so much for adding these bug fixes and warning messages - they're super helpful. I just made a couple of minor suggestions below. Otherwise, this is good to go.

wankiewiczm-smh:
Thanks for the review @loffleraSMH! I updated the queries to drop the explicit removal of paeds-only hospitals, now it relies on the `derived_variables` flags. I also set up the temp table for the warning message and the function still runs as expected.

Let me know if you'd like to test it one more time since the code is slightly different or if I'm okay to merge.

loffleraSMH:
> Thanks for the review @loffleraSMH! I updated the queries to drop the explicit removal of paeds-only hospitals, now it relies on the `derived_variables` flags. I also set up the temp table for the warning message and the function still runs as expected.
> 
> Let me know if you'd like to test it one more time since the code is slightly different or if I'm okay to merge.

Thanks so much @wankiewiczm-smh! I took a quick look at the additional changes and ran a quick test. Everything looks good to me, so feel free to go ahead and merge this.

